PR #1, opened 2019-03-31
Move Demo App to Repo

arjunsalyan:


.idea/encodings.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mojca: I would be in favour of removing everything inside `.idea`. Isn't this something that's specific to your editor?

MacPorts/settings.py
```diff
@@ -0,0 +1,139 @@
...
+# Build paths inside the project like this: os.path.join(BASE_DIR, ...)
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+TEMPLATE_DIR = os.path.join(BASE_DIR, 'templates')
+
+
+# Quick-start development settings - unsuitable for production
+# See https://docs.djangoproject.com/en/2.1/howto/deployment/checklist/
+
+# SECURITY WARNING: keep the secret key used in production secret!
+SECRET_KEY = 'v3@ml!wg8ssflyog$2b6am3$t)v!xu&e+0o=*j+0rln^+-4jv)'
```
mojca: Like it says: keep the secret key secret.

categories.json
```diff
@@ -0,0 +1,639 @@
+[
+  {
+    "model": "ports.Category",
+    "pk": "amusements",
+    "fields": {
+      "name": "amusements"
```
mojca: This file should go. You should not be hardcoding categories anywhere. You need to create categories dynamically while importing the ports database.

maintainer_fixtures.json
```diff
@@ -0,0 +1,5252 @@
+[
+    {
+        "model": "ports.Maintainer",
+        "pk": 0,
+        "fields": {
+            "github": "ryandesign",
```
mojca: This file should go. Maintainers should be read and dynamically populated / updated from portindex json data.

parsing-scripts/create_maintainers_fixtures.py
```diff
@@ -0,0 +1,103 @@
+import json
+
+with open('portindex.json', "r") as file:
+    data = json.load(file)
```
mojca: You should dynamically populate maintainers while reading portindex json. Read a port, check if maintainer exists; create one if it doesn't exist; else just find the maintainer's id and enter the relation port-maintainer in the additional table.

parsing-scripts/create_maintainers_fixtures.py
```diff
@@ -0,0 +1,103 @@
...
+            maintainer_object['name'] = name
+            maintainer_object['domain'] = domain
+            maintainer_object['github'] = github
+            maintainer_object['inconsistency'] = True
+
+        # Fifth Case, Github is repeated and Email is not provided (skip)
+        elif github_repeat and email_provided is False:
+            continue
+
+        # Sixth Case, Email is repeated and Github is not provided (skip)
```
mojca: This whole file sounds a lot more complicate than it could be if my first idea was implemented. Maybe this needs further discussion.

parsing-scripts/parse_build_history.py
```diff
@@ -0,0 +1,124 @@
...
+import urllib.request
+import ssl
+import json
+import datetime
+
+
+builders = [
+    '10.14_x86_64',
+    '10.13_x86_64',
+    '10.12_x86_64',
```
mojca: If these values are hardcoded, they should probably be hardcoded in some separate settings file, to avoid potentially repeating yourself in every script where you might need this list.

parsing-scripts/portindex2fixtures.py
```diff
@@ -0,0 +1,31 @@
...
+
+    newdata = {
+        'model': 'ports.Port',
+        'pk': i,
+        'fields': port
+    }
+    port_fixtures.append(newdata)
+    i = i + 1
+
+with open('../port_fixtures.json', 'w') as newjson:
```
mojca: Just curious: Why do you need to do this conversion? Couldn't this go to the database directly?

Anyway, please remove port_fixtures.json, that one should not be part of repository, it's simply too big and always outdated.

ports/migrations/0001_initial.py
```diff
@@ -0,0 +1,79 @@
...
+
+    dependencies = [
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='BuildHistory',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('builder_name', models.CharField(db_index=True, max_length=50)),
```
mojca, line 19: This needs to be a foreign key, referencing the database of builders.

ports/migrations/0001_initial.py
```diff
@@ -0,0 +1,79 @@
...
+                ('build_id', models.IntegerField()),
```
mojca, line 20: Indexed, please.

ports/migrations/0001_initial.py
```diff
@@ -0,0 +1,79 @@
...
+                ('status', models.CharField(max_length=50)),
```
mojca, line 21: Indexed.

ports/migrations/0001_initial.py
```diff
@@ -0,0 +1,79 @@
...
+                ('port_name', models.CharField(db_index=True, max_length=50)),
+                ('time_start', models.CharField(max_length=50)),
+                ('time_elapsed', models.CharField(max_length=50)),
```
mojca, line 24: These two fields need to represent a real timestamp rather than a string, so that you can potentially sort, do arithmetics, do queries about events that happened on a particular day, ...

ports/migrations/0001_initial.py
```diff
@@ -0,0 +1,79 @@
...
+                ('build_url', models.URLField(default='')),
+                ('watcher_id', models.IntegerField()),
```
mojca, line 26: indexed

ports/migrations/0001_initial.py
```diff
@@ -0,0 +1,79 @@
...
+                ('watcher_url', models.URLField()),
+            ],
+        ),
+        migrations.CreateModel(
+            name='Category',
+            fields=[
+                ('name', models.TextField(primary_key=True, serialize=False)),
```
mojca: Primary key could potentially be a number to save space, even though this is also an acceptable way.

ports/migrations/0001_initial.py
```diff
@@ -0,0 +1,79 @@
...
+            ],
+        ),
+        migrations.CreateModel(
+            name='Maintainer',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('name', models.CharField(max_length=50)),
```
mojca: indexed

ports/migrations/0001_initial.py
```diff
@@ -0,0 +1,79 @@
...
+                ('domain', models.CharField(max_length=50)),
```
mojca: indexed

ports/migrations/0001_initial.py
```diff
@@ -0,0 +1,79 @@
...
+                ('github', models.CharField(db_index=True, max_length=50)),
+            ],
+        ),
+        migrations.CreateModel(
+            name='Port',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('portdir', models.CharField(max_length=100)),
```
mojca, line 56: indexed

ports/migrations/0001_initial.py
```diff
@@ -0,0 +1,79 @@
...
+                ('variants', models.TextField(default='')),
```
mojca, line 57: In the future we require a better representation, in particular with respect to statistics collection. The oversimplification may stay for now.

ports/migrations/0001_initial.py
```diff
@@ -0,0 +1,79 @@
...
+                ('description', models.TextField(default='')),
+                ('homepage', models.URLField(default='')),
+                ('epoch', models.TextField(default='')),
```
mojca, line 60: this is number

ports/migrations/0001_initial.py
```diff
@@ -0,0 +1,79 @@
...
+                ('platforms', models.TextField(default='')),
```
mojca, line 61: not ideal, but may stay for now

ports/migrations/0001_initial.py
```diff
@@ -0,0 +1,79 @@
...
+                ('long_description', models.TextField(default='')),
+                ('depends_extract', models.TextField(default='')),
```
mojca: This eventually needs to reference another port, and absolutely needs to be indexed. It probably needs to go to a separate table. Same goes for all dependencies. We want to be able to tell which port depends on what.

ports/migrations/0001_initial.py
```diff
@@ -0,0 +1,79 @@
...
+                ('version', models.CharField(max_length=100)),
+                ('revision', models.TextField(default='')),
```
mojca: number

ports/migrations/0001_initial.py
```diff
@@ -0,0 +1,79 @@
...
+                ('closedmaintainer', models.BooleanField()),
+                ('name', models.CharField(db_index=True, max_length=100)),
+                ('license', models.CharField(max_length=100)),
```
mojca, line 67: Licence handling needs to be improved in the future. A port may have different licences, so this is not fully sufficient. May stay for now.

ports/migrations/0001_initial.py
```diff
@@ -0,0 +1,79 @@
...
+                ('depends_lib', models.TextField(default='')),
+                ('depends_build', models.TextField(default='')),
+                ('installs_libs', models.TextField(default='')),
+                ('depends_fetch', models.TextField(default='')),
+                ('depends_run', models.TextField(default='')),
+                ('conflicts', models.TextField(default='')),
+                ('replaced_by', models.TextField(default='')),
+                ('depends_test', models.TextField(default='')),
+                ('depends_patch', models.TextField(default='')),
```
mojca: See above. These are references to other ports.

ports/migrations/0001_initial.py
```diff
@@ -0,0 +1,79 @@
...
+                ('subports', models.TextField(default='')),
```
mojca: I would probably misuse `portdir` to handle info about subports.

requirements.txt
```diff
@@ -0,0 +1,23 @@
+asn1crypto==0.24.0
+beautifulsoup4==4.7.1
+bs4==0.0.1
+certifi==2019.3.9
+cffi==1.12.2
+chardet==3.0.4
+cryptography==2.6.1
+DateTime==4.3
+Django==2.1.7
```
mojca, line 9: Should also be tested against 2.2.

mojca (commented):
First batch of review (more to come)

.idea/encodings.xml
arjunsalyan: Yes, thank you. It should be removed.

categories.json
arjunsalyan: I wanted to provide some initial data which someone who tries to run the app on their system can use to populate the database.

But a python script can also generate this file, and so it can be removed.

maintainer_fixtures.json
arjunsalyan: This file is dynamically generated from the `portindex` using the scripts inside `parsing-scripts`. But again here also, I kept one copy to have some initial data.

There will be two types of scripts: [1] those which generate initial data in the form of Django fixtures. Fixtures are very efficient way of populating database with initial data.

[2] After the initial data is in the database, a script that would take the difference of portindex will keep our database updated. I had such script earlier, but since now the approach is changed and we probably will use `portindex -c` or `portindex -diff` for this purpose, the new script is not ready yet.

**[EDIT]**: We could probably avoid these and make a single script that would read portindex and update all the tables simultaneously. It would be lot better then.

Thank You

parsing-scripts/create_maintainers_fixtures.py
arjunsalyan: Yes, I have changed it to do the same. I will update the pull request once I am done with today's schedule.

parsing-scripts/create_maintainers_fixtures.py
arjunsalyan: This is almost the same idea you suggested in the email. The further complications have arrived just because not all maintainers have provided their GitHub handles. Otherwise it would have been very simple.

parsing-scripts/parse_build_history.py
arjunsalyan: Yes, I will be creating a table of builders, which can be updated as newer versions of macOS get released.

parsing-scripts/portindex2fixtures.py
arjunsalyan: > Just curious: Why do you need to do this conversion? Couldn't this go to the database directly?

It could go. But django fixtures are more efficient and they handle relations very well [for inserting initial data]. Updating can then be done directly as the data won't be very large.

But I guess, we could avoid some complexity by skipping fixtures and going for entering data directly. Thank You

requirements.txt
arjunsalyan: Will do. Thanks

arjunsalyan:
I will make all model-related changes you suggested (for others I have replied on the respective comments). Will be going step-by-step now, beginning from parsing of maintainers, because we first need to finalise the data we are going to store. Then I will do the models. Or the sequence needs to be changed?

Thank You

mojca:
Yes, regarding maintainers there are two different approaches. The one you implemented is what I would have initially done myself. Keep one entry per person. Keep reading the maintainers from the port list and try to keep a somewhat consistent state; if discrepancy is found, mark that discrepancy.

But after some thinking and given our imperfect state of matters with lots of inconsistencies in the ports, I thought that the first approach I described in the email might be a lot more straightforward. That is: keep the triple (github, email_name, email_domain) 100% consistent with what's in the Portfile, and add unique identifier to the triple. (Side note: you could merge email name and domain into a single string if you wanted, but I don't know if there's any real benefit to that. The reason why it's not done in portindex is to avoid exposing explicit emails to stupid bots and spammers.)

So you could in principle end up with something like this for maintainer list:

| id | github | name | domain |
|----|-------|------|--------|
| 1 | alpha | first | example.com |
| 2 | alpha | second | example.com |
| 3 | beta | | |
| 4 | alpha | | |
| 5 | gamma | first | example.com |
| 6 | | third | example.com |

with empty cells being either empty strings or nil, whatever is easier to handle.

A port that lists maintainer `github: alpha, email: { name: second, domain: example.com }` would link to maintainer id 2. Another port with maintainer `github: alpha` and no email would link to maintainer id 4.

Now, you make a page for maintainer `@alpha` and do the database search for any maintainers where `github = alpha`, returning you maintainers 1, 2, 4, and you list all ports associated with any of those maintainers, but you also show a warning, saying that this particular maintainer is associated with three different entries, and eventually (not now) you could show which port belongs to which entry (or at least the number of ports with that representation of the maintainer).

Alternatively one might want to display ports belonging  to maintainer `example.com:first`, resulting in maintainer id 1 and 5, and then showing all ports belonging to any of those two maintainer ids.

Now, admittedly this gets somewhat non-trivial as far as database queries are concerned. Nothing unusual for database queries and probably the same kind of complexity required for nice handling of statistics, the reason why I suggested some reading about databases, learning different joins, views etc. This is where creating an appropriate view would be handy.

But on the other hand it spares you the troubles of handling various inconsistencies yourself, and potentially ending up in situation when you would not be able to correctly update information once it gets fixed in the portfiles.

That said, it's not that important to get this implemented in a perfect way at this very moment. So maybe I'll "remove my request" to get it fixed now. While it is important to some extent, work on ports is way more important than dealing with maintainers' madness.

arjunsalyan:
> Now, you make a page for maintainer `@alpha` and do the database search for any maintainers where `github = alpha`, returning you maintainers 1, 2, 4, and you list all ports associated with any of those maintainers, but you also show a warning, saying that this particular maintainer is associated with three different entries, and eventually (not now) you could show which port belongs to which entry (or at least the number of ports with that representation of the maintainer).

This method is probably much more practical. But I was unable to implement this because of ports with details like that of `id 6`. There is no GitHub handle, won't It be missed if we use GitHub as the query criteria? Sorry, if I am missing something.

> That said, it's not that important to get this implemented in a perfect way at this very moment. So maybe I'll "remove my request" to get it fixed now. While it is important to some extent, work on ports is way more important than dealing with maintainers' madness.

OK! Thank You. Next, I will re-write the populating script (bring all into one simple step) and bring the models in the perfect shape.

mojca:
> This method is probably much more practical. But I was unable to implement this because of ports with details like that of `id 6`. There is no GitHub handle, won't It be missed if we use GitHub as the query criteria? Sorry, if I am missing something.

That's why you would allow different queries: both for github and for emails.

On the email I suggested using something like `/maintainer/@githubhandle` to list all ports belonging to maintainer with a particular github handle, and still allow displaying `/maintainer/name__domain.com` (or some other obfuscation) which would search for maintainers with a particular email (it could also be `/maintainer/githubhandle` vs. something like `/maintainer/m/domain.com/name`; or `/maintainer/github/githubhandle` vs. something else, even if `/maintainer/email/<hashed version of domain>/name` or `/maintainer/email/<hashed-version-of-full-email>`; also at some point in the future allowing `/maintainer/name/First Last`). The only important part is not to expose the maintainer's email address in some form that's (too) easy to harvest emails from for web spiders. The second form would still be able to display the page for maintainer with id 6. Note that this is not the only option, there are certainly other ways.

By hashes I meant that you could for example calculate some hash (like md5 or any other random one) of the full email and store it in the database along with other entries for maintainer, and then join by those hashes.

arjunsalyan:
This is a doubt regarding `depends_build`, `depends_lib` etc. Should this go to the mailing list?

We have:
- depends_lib
- depends_build
- installs_libs
- depends_fetch
- depends_run
- conflicts
- replaced_by 
- depends_test
- depends_patch
- subports

1. The value is not always in the same format, e.g. "`depends_build" : ["port:pkgconfig","path:bin\/cmake:cmake"], ` . Does path need to be treated differently or I can simply extract "cmake" from it?
2. How do I deal with the value of `depends_lib`, e.g. ` "depends_lib" : ["lib:libz.1:zlib","port:xorg-xorgproto"]`. Here what does `"lib:libz.1:zlib"` indicate and what is the format it follows?
3. Do they all need to be shown separately on port-detail or all can come under one key: "Dependencies"?

Thank You

mojca:
These are "ugly" technical details, so we may just as well keep the discussion here, and anyone interested can still read this thread and comment on it.

The `path:` or `bin:` prefix means that other ports can satisfy the dependency as well. In case of `cmake` that's probably `cmake-devel`. In theory this is any port that contains the file `${prefix}/bin/cmake`. What I would do for now is treat `path:`-style dependencies in the same way as any other `port:`-style dependency. Optionally you could add another column to the table with dependencies that would be `nil` or empty by default (when dependency is specified as `port:foo`), while it would contain `path:bin/cmake` in the above example. In the first stage, you just store that value and don't do anything special with it. In later stages you could optionally mark such a special dependency with a different visual marker on the page. In some extension goals (really not that interesting at the moment) and when full list of files of all ports is available, one could list alternative ports that could satisfy the dependency :)

The format of `"lib:libz.1:zlib"` is `"lib:<any-filename-of-a-library>:<portname>"`. Store the portname as usual, as if it was `"port:zlib"`. Basically: anything following the last colon should be port name. if whatever comes before that last colon is not `"port"`, you could store that part in that extra optional column, in this case `"lib:libz.1"`. This means that any port providing library `libz.1` satisfies the dependency.

Regarding dependencies: you could probably put all dependencies into the same table, just keep one column to store the type of dependency (`lib`, `test`, `run`, ...). I don't care what precisely is shown on the website at this particular moment (it's not as high priority), but at the end we would want to display different types of dependencies separately.

`installs_libs` is a boolean value (probably wrong in most ports, something that could be fixed within the scope of this project, but no need to worry about it now). It tells you whether the port installs some libraries or not.

`conflicts` should go to a separate table. Unless there's a mistake in portfiles, if A conflicts with B, B also conflicts with A.

`replaced_by` should be part of the main table with ports.

`subports` don't need to be entered (and should not be if you want to keep your database design consistent). They can be identified by sharing the same `path`.

arjunsalyan:
Thank You.
While I was parsing depends_*, I found that while writing the dependencies, sometimes the case of the ports is not taken care of, example, for grass7, it is `port:py27-pillow` instead of `"port:py27-Pillow"`.
Shall I go ahead and try to fix such cases, or simply make the query case-insensitive. I am not aware of how much case matters in the naming of ports, if it is of importance then these cases should be fixed.

mojca:
MacPorts is generally case-insensitive as far as port names go (but not when it comes to variants, for example; that is considered a bug, but still). Nevertheless the case should match, and any such inconsistencies should ideally be reported and/or fixed. I don't have a super good answer with respect to what to do in the app when something like this happens. The app should not crash or fail to update. It would be nice to report such cases, but I would not turn the database design on its head just to be able to support such reporting. (This would be possible to do by using port name instead of port id as dependency reference, but I'm not sure if it's worth doing it.)

Regarding dependencies: one option would be to further improve `portindex2json` to print all dependencies inside a single key, for example:
```
{
    "name": "PortName",
    "depends": -- or "dependencies"
        {
            "fetch": ["bin:git:git"],
            "run": ["port:python"],
            "test": ["port:py-test"],
        }
}
```
We could further improve the `bin`/`port`/`path`/`lib` specification, but I'm somewhat afraid that the size of the file would explode too soon.

Regarding the bootstrapping (first run with the need to fill the database). It might in fact make sense to use a special more efficient format to initialize the database. But that file should not be in repository. Instead, the repository should contain either instructions or a script to prepare such files on the fly before feeding the database. (I assume one could also prepare such files in memory?) The first step is to get `PortIndex` from somewhere (copy the local one or fetch it from server), then run `pontindex2json`; finally a single python script should "populate" the database for all of the categories, maintainers, ports, dependencies etc. Generating those fixtures would count as a valid way of populating the database, but I would like to see it done with a single script (load json, iterate through the ports, generate categories and maintainers on the fly, store all info somewhere - either to a single file or to the database). Just not put the gazilion lines to the repository.

arjunsalyan:
> MacPorts is generally case-insensitive as far as port names go (but not when it comes to variants, for example; that is considered a bug, but still). Nevertheless the case should match, and any such inconsistencies should ideally be reported and/or fixed. I don't have a super good answer with respect to what to do in the app when something like this happens. The app should not crash or fail to update. It would be nice to report such cases, but I would not turn the database design on its head just to be able to support such reporting. (This would be possible to do by using port name instead of port id as dependency reference, but I'm not sure if it's worth doing it.)


We can achieve this without any changes to the database. Every time we encounter a DoesNotExist exception for any dependency we can store them separately and thus report them.

> Regarding the bootstrapping (first run with the need to fill the database). It might in fact make sense to use a special more efficient format to initialize the database. But that file should not be in repository. Instead, the repository should contain either instructions or a script to prepare such files on the fly before feeding the database. (I assume one could also prepare such files in memory?) The first step is to get `PortIndex` from somewhere (copy the local one or fetch it from server), then run `pontindex2json`; finally a single python script should "populate" the database for all of the categories, maintainers, ports, dependencies etc. Generating those fixtures would count as a valid way of populating the database, but I would like to see it done with a single script (load json, iterate through the ports, generate categories and maintainers on the fly, store all info somewhere - either to a single file or to the database). Just not put the gazilion lines to the repository.


I have created a single script that would load categories, ports and maintainers directly into the database without using fixtures. Initially the script took **6 minutes** to populate the entire portindex (categories, ports and maintainers) on a local database. But then I reduced the number of queries by storing the data to compare-with separately and not query the database each time. The time needed then went down to **2 min 53 sec** . It is still not very great. But it reduces the complexity of handling fixtures and the sequence in which they should be loaded. Both the methods are working fine though and we can go for any. For loading any database not hosted on the the same server- none of these is a good method given the large size. For loading any external database the best way may be using `sql dumps`.

Regarding loading the dependencies- it is the most time consuming of all processes. We cannot do this using fixtures because of the naming inconsistencies and if we load them directly into the database: I tried- but the script is currently doing many thousand queries and is really slow.

Please do review the files:
- `parsing-scripts/load_initial_data.py`      [please ignore the complexity in parsing of maintainers, I am still figuring out a lot of things, but do check if I am loosing on efficiency]
- `parsing-scripts/load_inital_depends.py`
- `ports/models.py`        [ `class Port` and `class Dependency`, `replaced_by` is still a CharField ]



> Regarding dependencies: one option would be to further improve `portindex2json` to print all dependencies inside a single key, for example:


Yes, even I felt this, and `conflicts` should also be a list of strings.

Thank You

mojca:
Unrelated to the above discussion ...

Here's a sample representation of one random package from FreeBSD: [link](https://www.freshports.org/lang/python27/), listing dependent ports, dependencies, variants with descriptions, the equivalent of `port notes`, ... I need to admit that I like Homebrew's pages more though.

ports/migrations/0001_initial.py
```diff
@@ -52,28 +60,29 @@ class Migration(migrations.Migration):
                 ('epoch', models.TextField(default='')),
                 ('platforms', models.TextField(default='')),
                 ('long_description', models.TextField(default='')),
-                ('depends_extract', models.TextField(default='')),
                 ('version', models.CharField(max_length=100)),
-                ('revision', models.TextField(default='')),
+                ('revision', models.IntegerField(default=2)),
```
mojca, line 64: Why `default=2`? If anything, it should be 0.

ports/models.py
```diff
@@ -1,21 +1,12 @@
 from django.db import models
 
 class Category(models.Model):
-    name = models.TextField(primary_key=True)
+    name = models.TextField(primary_key=True, db_index=True)
```
mojca: This is probably the one place where `db_index=True` is not needed. Primary key is automatically indexed.

ports/models.py
```diff
@@ -27,23 +18,31 @@ class Port(models.Model):
     platforms = models.TextField(default='')
     categories = models.ManyToManyField(Category, related_name='category', db_index=True)
     long_description = models.TextField(default='')
-    depends_extract = models.TextField(default='')
     version = models.CharField(max_length=100)
-    revision = models.TextField(default='')
-    maintainers = models.ManyToManyField(Maintainer, related_name='maintainer')
+    revision = models.IntegerField(default=2)
```
mojca: Why default of 2?

ports/models.py
```diff
@@ -27,23 +18,31 @@ class Port(models.Model):
...
-    depends_build = models.TextField(default='')
-    installs_libs = models.TextField(default='')
-    depends_fetch = models.TextField(default='')
-    depends_run = models.TextField(default='')
-    conflicts = models.TextField(default='')
-    replaced_by = models.TextField(default='')
-    depends_test = models.TextField(default='')
-    depends_patch = models.TextField(default='')
-    subports = models.TextField(default='')
+    installs_libs = models.BooleanField(default=False)
```
mojca: I believe our implicit default is True, but I'm not actually sure. (To be honest ... I might be inclined to skip this for now until something useful is implemented with it.)

ports/models.py
```diff
@@ -27,23 +18,31 @@ class Port(models.Model):
...
+    port_name = models.ForeignKey(Port, on_delete=models.CASCADE, related_name="dependent_port", db_index=True)
+    dependencies = models.ManyToManyField(Port, db_index=True)
+    type = models.CharField(max_length=100)
+
+
+class Maintainer(models.Model):
+    name = models.CharField(max_length=50, db_index=True)
+    domain = models.CharField(max_length=50, db_index=True)
+    github = models.CharField(max_length=50, db_index=True)
+    ports = models.ManyToManyField(Port, related_name='maintainer_port', db_index=True)
```
mojca: Shouldn't `related_name` better be `maintainers` here?

ports/models.py
```diff
@@ -27,23 +18,31 @@ class Port(models.Model):
...
+
+    objects = PortManager()
+
 
 class BuildHistory(models.Model):
     builder_name = models.CharField(max_length=50, db_index=True)
```
mojca: Foreign key (referencing) builders.

ports/models.py
```diff
@@ -27,23 +18,31 @@ class Port(models.Model):
...
+    installs_libs = models.BooleanField(default=False)
+    replaced_by = models.CharField(max_length=100)
+
+    objects = PortManager()
+
+
+class Dependency(models.Model):
+    port_name = models.ForeignKey(Port, on_delete=models.CASCADE, related_name="dependent_port", db_index=True)
+    dependencies = models.ManyToManyField(Port, db_index=True)
+    type = models.CharField(max_length=100)
```
mojca, line 35: This field should be indexed as well. And for the sake of consistency ideally referencing (with a foreign key) yet another table holding types of dependencies.

parsing-scripts/load_initial_data.py
```diff
@@ -0,0 +1,191 @@
...
+        pass
+    try:
+        new_port.replaced_by = port['replaced_by']
+    except KeyError:
+        pass
+    new_port.save()
+    try:
+        new_port.categories.add(*port['categories'])
+    except KeyError:
+        pass
```
mojca: I assume you could use one try and one except here?

mojca (commented):
Just some nitpicking. I need to check the python scripts again in more detail and maybe finally do some local testing. (It's nothing wrong if you store the data locally first before pushing it to the database.)

arjunsalyan:
> Unrelated to the above discussion ...
> 
> Here's a sample representation of one random package from FreeBSD: [link](https://www.freshports.org/lang/python27/), listing dependent ports, dependencies, variants with descriptions, the equivalent of `port notes`, ... I need to admit that I like Homebrew's pages more though.

Thanks, it is really helpful. I quickly implemented dependents and dependencies which are both working fine locally (both database and app on local server) but with my AWS Database it won't just load. The query to find `dependents` is causing the trouble.

views.py
```
def portdetail(request, name):
    port = Port.objects.get(name=name)
    maintainers = Maintainer.objects.filter(ports__name=name)
    dependencies = Dependency.objects.filter(port_name_id=port.id)
    dependents = Dependency.objects.filter(dependencies__id=port.id)
```

port-detail.html
```
<tr>
  <th scope="row">{{ dependents|length }} Port(s) that depend on {{ port.name }}:</th>
    <td>
      {% for dependent in dependents %}
         <a href="{% url 'port_detail' dependent.port_name.name %}">{{ dependent.port_name.name }}</a> |
      {% endfor %}
  </td>
</tr>
```

With this addition the response time has increased very heavily. Works totally fine local database but with AWS it takes like minutes to load and Heroku times out the builder at 30sec.
Removing the `for loop` from `port-detail.html` or the `dependent` query altogether makes it fine again.

Also, for the ports which have zero dependents, it loads fine.

[AppHack](https://frozen-falls-98471.herokuapp.com/ports/AppHack/) : Loads fine, has zero dependents
[unzip](https://frozen-falls-98471.herokuapp.com/ports/unzip/): Crashes

mojca:
As a quick fix just comment that query out for now and we can take a look later to avoid being stuck on a stupid problem that's not even relevant for now. Database were made for such scenarios, the query should be fast, there might be just some minor issue in the code (not yet sure where).

mojca:
One general question. You added more than 500 files to the repo, most of which are static / javascript files, generated or copied from somewhere. Is there some alternative way to get those files? I'm not saying there must be one, I just don't know how to take care of updates when they would be needed etc. I know that in case of `npm` modules one would just list which modules to install and then the user would first run `npm` to fetch the gazillion of files (also making updates easier to manage).

arjunsalyan:
> One general question. You added more than 500 files to the repo, most of which are static / javascript files, generated or copied from somewhere. Is there some alternative way to get those files? I'm not saying there must be one, I just don't know how to take care of updates when they would be needed etc. I know that in case of `npm` modules one would just list which modules to install and then the user would first run `npm` to fetch the gazillion of files (also making updates easier to manage).

Yes, I checked, those files are generated by django itself as a part of the ready-made admin section that comes with it. These files can be totally removed from the repo, in fact, the `staticfiles` directory can be altogether removed because it can be generated by the `collectstatic` command.

**UPDATE**: Thank You, we could get rid of 477 useless files.

parsing-scripts/load_initial_data.py
arjunsalyan: But then if any one of the keys is not present, won't it stop adding others also?

ports/models.py
arjunsalyan: Fine, I will remove it for now.

ports/models.py
arjunsalyan: Sorry, just a stupid mistake!

ports/models.py
arjunsalyan: Thank You

mojca:
> **UPDATE**: Thank You, we could get rid of 477 useless files.

Awesome, thank you. This was one of my major concerns. Doing a code review with so many files was nearly impossible. Sure, I knew I should ignore the static files, but the page listing the diffs was way to long and I could have missed the important stuff. This now got to a manageable state. Yay!

parsing-scripts/load_initial_data.py
mojca: OK, sorry, I missed this detail. Now, I would then do it in a slightly different way. There are some keys which are absolutely necessary (port name, port version, port dir, ...), others which are optional (a port may have no variants, no dependencies, no maintainers, no licence [not quite ok, but possible scenario], ...) and some with defaults (if there is no revision or no epoch, the implied value is 0; if the "closedmaintainer" is missing, the value should be false, ...).

Instead of catching an exception while reading a non-existing key, you could instead check if the key exists and act accordingly, avoiding the exception. You should spare exception handling for other strange scenarios like ending up in invalid database entry, too long name etc. 

I would start with important / necessary fields first (port name, ...) and if an error happens there, you should probably print the error and maybe not add the port at all? (If there's a serious error there, I would expect further errors down the road, so it's not as important what happens next.) For other fields you should still catch errors, but doing it once would probably be sufficient. However make sure that you are not reading invalid keys / that you replace them with some default value.

arjunsalyan:
I was going through Django Documentation and came across "Postgres Full Text Search". I tried to implement it in the demo app for "search by description". But it doesn't really work well. Seems like it searches only when a complete word is typed and not for partial.

Example:
It does not show any results for "manipulat" and returns many results for "manipulate"
[Demo App](https://frozen-falls-98471.herokuapp.com)

parsing-scripts/load_initial_data.py
arjunsalyan: Thank You

mojca:
Cool, thank you very much!

The following looks like a helpful [blog entry](https://medium.com/@pauloxnet/full-text-search-in-django-with-postgresql-4e3584dee4ae).

I didn't try it myself, but your search for port name already works well and does partial matches.

The usual SQL keyword for matching part of the field is `LIKE` and doesn't care if it's a full or just a partial word match.

(Without testing) I believe that `__search` vs. `__icontains` is doing precisely what you (maybe?) didn't want it to do (or at least what you complained about): only show you full word matching (`__icontains` looks like case insensitive match, probably what you do want to use). Now, this is a bit of a tricky business. You might want to give priority to full word match, but also show partial matches at the end. Partial matching is often undesired, which is likely the reason why it was implemented this way, but it's probably helpful when there are no other hits.

Again, I'm not yet sure about what we want. Full match first, partial match last sounds like a reasonable choice, but there is no need to do science out of some search box at this point. I don't care if we don't do any further modifications now.

arjunsalyan:
> The following looks like a helpful [blog entry](https://medium.com/@pauloxnet/full-text-search-in-django-with-postgresql-4e3584dee4ae).

Thank You

> Again, I'm not yet sure about what we want. Full match first, partial match last sounds like a reasonable choice, but there is no need to do science out of some search box at this point. I don't care if we don't do any further modifications now.

Yes, it also does the job for now. Our priority at this point could be keeping the PortIndex updated (from loading initial data to handling changes- `parsing-scripts`) and the related models.

Regarding the port-detail page: Won't it be useful if we have an `updated_at` field and also show something like when was the data for that port last synced with the portindex?

arjunsalyan:
> (Without testing) I believe that `__search` vs. `__icontains` is doing precisely what you (maybe?) didn't want it to do (or at least what you complained about): only show you full word matching (`__icontains` looks like case insensitive match, probably what you do want to use). Now, this is a bit of a tricky business. You might want to give priority to full word match, but also show partial matches at the end. Partial matching is often undesired, which is likely the reason why it was implemented this way, but it's probably helpful when there are no other hits.

Yes, `__icontains` is case insensitive because of that "i". In past what I used to do was split the `search__text` and for each splitted word use `__icontains` and query the database several times. The results could then be merged retaining the original ranking given by `__icontains`, and it did well. But again the data set was very small compared to what we have now.

arjunsalyan:
I have added separate table for the builders. And we could get rid of hardcoding the builder names anywhere in the app. Also, I have made a slight change on port-detail: now it shows "No Build History Available for this builder", instead of nothing, whenever the history is not available.
Example: [libxo](https://frozen-falls-98471.herokuapp.com/ports/libxo/) & [AppHack](https://frozen-falls-98471.herokuapp.com/ports/AppHack/)

mojca:
Great! The `No Build History Available for this Builder` should be some light grey shade though rather than red. The red colour should be reserved for build failures.

One general request. I'm currently travelling and wanted to give the app some more testing, and play with it a bit.

Could you please also update your installation instructions which were supposed to change after you cleaned up the bootstrapping process? (I'm collecting my own notes about how to make it run with an existing MacPorts installation on local computer. We might want to document usage on, say, Debian and perhaps set up a docker image later on.)

The latest commit looks fine.

The following is a random idea which doesn't have a super high priority, but give that you have some data now, it might not be too much work to implement (I still need to provide more feedback about the table design.) It would be nice to prepare a view with all the latest builds from the buildbot (it could be a paginated view, or simply limited to a fixed number as the first approximation) and a filter to select between "all builds", "successful builds", "broken builds". (There could be a second filter to select just one builder.) Again: not super important, it would just allow us to get some better quick insight into the build system.

Thanks again.

arjunsalyan:
I have updated the instructions.

'All Builds' for sure would be a very useful page, I have started working on it. It is just that my pace would be a bit slow (exams 😓 , but just two left 😃).

arjunsalyan:
I have added the [all builds](https://frozen-falls-98471.herokuapp.com/all_builds/) view.

The page right now contains a simple (and only one) filter. Adding the functionality to apply more than one filter simultaneously would need some more time.

I have decided in total these filters on the all builds page (which can be applied simultaneously):
- Builder
- Build Status
- Date
- Port ? (not sure about this)

**UPDATE:** `elapsed time` for failed builds is behaving strangely because it is "-1". It should better be left blank?

arjunsalyan:
I locally tested receiving submissions from mpstats and it worked fine. The base (users) table, which will be used to populate tables of port installations, os versions etc. is being updated/ added correctly.

The post request made by mpstats is:
```
submissions[data]={
....
}
```
which could not be directly read by `json.loads`, so I am using that ugly workaround of splitting at "=". Maybe we will have to make mpstats send just the JSON object and not "submissions[data]".

Now I am looking for some way to update the other tables as soon as this object is received from mpstats without affecting the response time (something like queuing the requests as in node.js).

Also, for "version" fields we can use `models.DecimalField`, but I was afraid that someday they might add letters to the version?

arjunsalyan:
I have made the changes to store all submissions instead of just the latest one. Also I have deployed it now on Heroku: [submissions](https://frozen-falls-98471.herokuapp.com/statistics/submit/) which is working correctly.

ports/tests/test_maintainers.py
```diff
@@ -0,0 +1,15 @@
...
+
+class TestMaintainers(TestCase):
+    def test_unique_entries_created(self):
+        ports = load_initial_data.open_portindex_json("ports/tests/testindex.json")
+        load_initial_data.load_categories_table(ports)
+        load_initial_data.load_ports_and_maintainers_table(ports)
+        self.assertEquals(Maintainer.objects.count(), 5, "Failed to create unique entities for maintainers")
+
+
+
```
mojca, line 32: Nitpicking: too many empty lines at the end :)

ports/tests/testindex.json
```diff
@@ -0,0 +1,101 @@
...
+            "domain" : "email.com",
+            "name"   : "user"
+        },
+        "github" : "user2"
+    }],
+    "long_description" : "Just a test port written to test something.",
+    "categories"       : ["categoryA"],
+    "version"          : "1.0.0",
+    "revision"         : "0"
+}]
```
mojca: Nitpicking: line ending (one empty line at the end) missing.

That said, for the real feedback:
* You'll need multiple such files for testing, so you should better name the file in a way that with make it easy to add further tests and still identify what this file was for (for testing maintainer import).
* I would like to see the `Portfile`s here as the source, as that's much easier and less error prone to edit. Then those `Portfile`s should be converted to some index of course. I'm not sure about the best way to do it in an elegant way (`Makefile` would be an option; not super elegant, but one way of doing the job; another option would be a simple bash or python script). Of course you would store both `Portfile`s and the json file to repository, but it should be easy (and a single command) to regenerate all the test json files from `Portfile`s.

mojca (commented):
Thank you very much, the first unit test generally looks ok to me. (I would need to test it and see how the database gets populated.)

ports/tests/testindex.json
arjunsalyan: > * I would like to see the `Portfile`s here as the source, as that's much easier and less error prone to edit. Then those `Portfile`s should be converted to some index of course. I'm not sure about the best way to do it in an elegant way (`Makefile` would be an option; not super elegant, but one way of doing the job; another option would be a simple bash or python script). Of course you would store both `Portfile`s and the json file to repository, but it should be easy (and a single command) to regenerate all the test json files from `Portfile`s.

Okay. I just realised that automating the entire process requires `portindex2json.tcl` to output a file, while it currently just prints the entire JSON. Shall I make a PR to make `portindex2json.tcl` write to a file `portindex.json` or the web app should maintain its own separate copy? I am not sure if writing to a file will make any difference, and if it does not, then it makes sense to make a PR.

MacPorts/__init__.py
```diff
@@ -0,0 +1,8 @@
+import os
+
+
+def setup():
+    module = os.path.split(os.path.dirname(__file__))[-1]
+    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "{}.settings".format(module))
+    import django
+    django.setup()
```
mojca: We usually want an empty line at the end.

MacPorts/urls.py
```diff
@@ -0,0 +1,32 @@
...
+Class-based views
+    1. Add an import:  from other_app.views import Home
+    2. Add a URL to urlpatterns:  path('', Home.as_view(), name='home')
+Including another URLconf
+    1. Import the include() function: from django.urls import include, path
+    2. Add a URL to urlpatterns:  path('blog/', include('blog.urls'))
+"""
+from django.contrib import admin
+from django.urls import path
+from django.urls import include
```
mojca: These two lines can be combined into a single one.

README.md
```diff
@@ -1,4 +1,45 @@
...
+```
+```
+python manage.py runserver
+```
+ 
+ 
+ 
+
+
+
```
mojca: ... just a single empty line :)

parsing_scripts/load_initial_data.py
```diff
@@ -0,0 +1,122 @@
...
+        try:
+            for maintainer in port['maintainers']:
+                try:
+                    name = maintainer['email']['name'].lower()
+                    domain = maintainer['email']['domain'].lower()
+                    email_provided = True
+                except KeyError:
+                    name = None
+                    domain = None
+                    email_provided = False
```
mojca: This looks ugly. Instead you should check whether the key exists. I assume the syntax is something like
```python
name = None
domain = None
github = None

if maintianer.has_key('email'):
    if maintainer['email'].has_key('name'):
        name = maintainer['email']['name'].lower()
    if maintainer['email'].has_key('domain'):
        domain = maintainer['email']['domain'].lower()
if maintainer.has_key('github'):
    github = maintainer['github'].lower()
```
That said, I'm not sure if I would downcase github handles. You can probably provide
```python
Maintainer.objects.get(..., github__iexact=github)
```
or something like that.

parsing_scripts/load_initial_data.py
```diff
@@ -0,0 +1,122 @@
...
+
+                try:
+                    github = maintainer['github'].lower()
+                    github_provided = True
+                except KeyError:
+                    github = None
+                    github_provided = False
+
+                # Check if the maintainer already exists in Database:
+                if email_provided is True and github_provided is True:
```
mojca: All this if-else-ing is strange. The database should be able to find the exact maintainer even if github handle is empty, as long as you make it consistent (either null or empty string; I guess null in this case).

parsing_scripts/load_initial_data.py
```diff
@@ -0,0 +1,122 @@
...
+                    except ObjectDoesNotExist:
+                        pass
+
+                # If the maintainer was not found in the database already, then only this code will run:
+                new_maintainer = Maintainer()
+                new_maintainer.name = name
+                new_maintainer.domain = domain
+                new_maintainer.github = github
+                new_maintainer.save()
+                new_maintainer.ports.add(new_port)
```
mojca: Wouldn't the `get_or_create()` method work here?
https://docs.djangoproject.com/en/2.2/ref/models/querysets/#get-or-create

parsing_scripts/load_initial_depends.py
```diff
@@ -0,0 +1,69 @@
...
+
+i = 0
+fixtures = []
+for port in ports:
+    iport = Port.objects.get(name__iexact=port['name'])
+    # Update Dependencies for the port
+    try:
+        load_depends(port['depends_lib'], "lib", iport)
+    except KeyError:
+        pass
```
mojca: You should check whether the key exists instead of catching exceptions here. Or, even better, since you have the function already, why not pass the full port to the function, and then the function checks whether that key exists.

You could easily call the function simply as
```python
load_depends(port, 'lib', iport)
```
instead of all the four lines combined.

Further, you could actually squeeze the full code into two lines, with something like
```python
for dependency_type in ['lib', 'extract', 'run', ...]:
    load_depends(port, dependency_type, iport)
```

parsing_scripts/load_initial_depends.py
```diff
@@ -0,0 +1,69 @@
...
+    dependency.dependencies.add(*dependencies)
+
+
+with open('portindex.json', "r") as file:
+    ports = json.load(file)
+
+i = 0
+fixtures = []
+for port in ports:
+    iport = Port.objects.get(name__iexact=port['name'])
```
mojca: Can we find a better name? The `iport` doesn't tell me anything and makes it somewhat more difficult to understand the code.

Btw: what happens if the database fails to find that port?

ports/tests/testindex.json
mojca: > Okay. I just realised that automating the entire process requires `portindex2json.tcl` to output a file, while it currently just prints the entire JSON.

You can pipe the result.
```bash
portindex2json.tcl {whatever parameters are needed} > portindex.json
```
This should create a file called `portindex.json`.

Sure, we could have a flag that optionally creates a file, and if that flag is missing, outputs everything to stdout.

> Shall I make a PR to make `portindex2json.tcl` write to a file `portindex.json`

You may. But the filename should not be hardcoded (it should be a parameter), and it should not be a showstopper as piping output should still work for your purpose. Or did you experience some issues with that?

> or the web app should maintain its own separate copy?

Ideally not. If you need that script, it's better to clone it / fetch it from the official repository. If you urgently need some changes which have not yet made it into the official version, you can clone from a different repository or branch as a temporary workaround, but if you urgently need something, please complain loudly :)

arjunsalyan:
I tried to use bfg to clean the `settings.py` file to remove `SECRET_KEY`, but it is not cleaning it from GitHub. I will look more into it.

I have improved the script `load_initial_data.py`, it is much smaller now. Also, Django does not recommend using `null=True` on string fields so I changed them to be empty strings.

I have added a command `python manage.py load` that will load the initial data. We can have such commands for `update`, `build-history`.

I will now work on `load_initial_depends.py`. It can be merged with `load_initial_data.py`.

ports/tests/testindex.json
arjunsalyan: > ```shell
> portindex2json.tcl {whatever parameters are needed} > portindex.json
> ```

Thank You so much. Can't believe I totally missed it. This solves everything.

parsing_scripts/load_initial_data.py
arjunsalyan: > ```python
> name = None
> domain = None
> github = None
> 
> if maintianer.has_key('email'):
>     if maintainer['email'].has_key('name'):
>         name = maintainer['email']['name'].lower()
>     if maintainer['email'].has_key('domain'):
>         domain = maintainer['email']['domain'].lower()
> if maintainer.has_key('github'):
>     github = maintainer['github'].lower()
> ```

Thank You, using `get()` made it even simpler.

parsing_scripts/update.py
```diff
@@ -0,0 +1,66 @@
...
+
+def open_portindex_json(path='portindex.json'):
+    with open(path, "r") as file:
+        ports = json.load(file)
+    return ports
+
+
+def full_update_ports(ports):
+
+    Maintainer.objects.all().delete()
```
mojca: Deleting all maintainers sounds too extreme. Maybe we could have some brainstorming together (voice or real-time written chat or something).

ports/tests/sample_data/category B/port-B1/Portfile
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           xcode 1.0
+version             1.0.0
+
+name                port-B1
+categories          categoryB
```
mojca: You seem to be using spaces in folder name. That almost sounds like a bad idea, and is probably also inconsistent with category (and `port lint` would usually complain about it as well).

ports/tests/sample_data/category B/port-B1/Portfile
```diff
@@ -0,0 +1,24 @@
...
+license             MIT
+
+description         This is port B1 of categoryB
+long_description \
+                    Just a test port written to test something.
+
+homepage            http://portB1.website/
+
+checksums           rmd160  1846701bae455bd934efe6dc025032f1fc47211e \
+                    sha256  5ed3f3d14d90c6c68980acb688c2d2a44a2eb78581a420951da7861a935db18d
```
mojca: Does macports complain if you leave the imaginary checksums out?

ports/tests/sample_data/category B/port-B1/Portfile
```diff
@@ -0,0 +1,24 @@
...
+
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           xcode 1.0
+version             1.0.0
+
+name                port-B1
+categories          categoryB
+platforms           darwin
+maintainers         {@user}
```
mojca: Doesn't hurt as a test, but single entry doesn't need parenthesis.

ports/tests/sample_data/category B/port-B1/Portfile
```diff
@@ -0,0 +1,24 @@
...
+license             MIT
+
+description         This is port B1 of categoryB
+long_description \
```
mojca: No need for a newline.

parsing_scripts/update.py
arjunsalyan: Yes, even I am looking for other ways of doing this. But updating the maintainers table seems really difficult because we have treated any combination of email and GitHub as a single unique entity. And hence, we cannot find the previous entry and change it.

But there is one more way I could think of:
1. While updating the info of a given port clear all the relations that exist with the maintainers table.
2. Add new maintainers, and new relations.
3. At the end if there happens to be any maintainer in the maintainers table with no ports associated with it -> delete it.

But the same purpose is served by first deleting them all, maybe?

We can do brainstorming as per your availability. Voice and written chat both work for me.

mojca: I mean ... deleting maintainers themselves seems a bit of an overkill. Also imagine that at some point in the future you'll add additional information like real name, password hash, links to trac tickets etc. Deleting relations between the ports and maintainers is less invasive, but deleting all the maintainers is somewhat problematic. Your approach described above seems better.

I would need to check whether Django supports database transactions. Port update should be a single transaction (if supported; not a big problem if we don't do it as a single transaction for now) which would delete people who are no longer maintainers and add new ones. You only add a new maintainer to the maintainers table if that maintainer doesn't exist already, and then only update the relations between the two.

You could in theory delete all relations at once, but note that if this in fact takes fourty minutes on heroku, the whole webpage will be useless during those fourty minutes.

It would be nice to have an action that takes a list of ports to update. The list could either be ten or thousand ports listed by name, or some kind of signal to say that we should update all ports. I would say that the procedure for updating should be almost the same in both cases, and in any case this is relatively agnostic as to whether it runs on inefficient server or somewhere where all actions are super fast.

arjunsalyan: > I mean ... deleting maintainers themselves seems a bit of an overkill. Also imagine that at some point in the future you'll add additional information like real name, password hash, links to trac tickets etc. Deleting relations between the ports and maintainers is less invasive, but deleting all the maintainers is somewhat problematic. Your approach described above seems better.

Okay, we can also avoid deleting all the relations for one port at once (as discussed in the alternate approach). But the queries are increasing with every bit of complication we add.

> You could in theory delete all relations at once, but note that if this in fact takes fourty minutes on heroku, the whole webpage will be useless during those fourty minutes.

Hmm, that is an important point I missed.

> It would be nice to have an action that takes a list of ports to update. The list could either be ten or thousand ports listed by name, or some kind of signal to say that we should update all ports. I would say that the procedure for updating should be almost the same in both cases, and in any case this is relatively agnostic as to whether it runs on inefficient server or somewhere where all actions are super fast.

Like a POST request with the list of ports? Or just some python command?
If I am getting it right, we take a list of names of ports and pick only those ports from the JSON file- and then we hand over the JSON data to our same `update.py` script?

mojca: It might be ok to delete the relations between ports and maintainers for one port at once (not the maintainers, just the relations). But it would be super strange to remove all relations at the same time (or worse, all maintainers).

You probably want to implement some subsection of the following (not necessarily all):
* command that updates all ports (and uses the local copy of portindex json)
* command that updates the given list of ports (and uses portindex json)
* command that updates ports from a given json object (and that object contains all data about all ports that you care about updating / adding / removing)

Once you have that functionality in place, there are different alternatives about how you want to run them. Yes, one option would be a post request (you just need to make sure that you don't end up with the whole world being able to spam the applications with requests to update random ports). Another option would be a local script that triggers the update (there could be a local job inspecting git updates, running portindex, portindex2json and making requests for updates). The basic update functionality should not care about how the updates are triggered. It's just the triggers on top that govern this. We could decide to trigger via post actions from the buildbot, or from local macports installation, or something else. As long as it is secure, efficient, reliable, easy to understand, ... I don't care as much how this triggering part is done.

arjunsalyan: > It might be ok to delete the relations between ports and maintainers for one port at once (not the maintainers, just the relations). But it would be super strange to remove all relations at the same time (or worse, all maintainers).

Okay, the latest commit removes only the relations between ports and maintainers for one port at once. All, we will have to do next is take care of the maintainers with zero ports related to them, which is unlikely but if someone tries to get rid of the inconsistency reported by the app- then it is very essential.

> * command that updates all ports (and uses the local copy of portindex json)

We have this.

> * command that updates the given list of ports (and uses portindex json)

Will add this.

> * command that updates ports from a given json object (and that object contains all data about all ports that you care about updating / adding / removing)

We also have this. The functions that update the database just receive a list of JSON objects. So, for full update, we pass it the entire JSON read from local portindex.json, for partial update we pass to it the selected JSON objects and for the third one we pass to it the JSON received.

ports/views.py
```diff
@@ -0,0 +1,293 @@
...
+        'all_builds': builds,
+        'filter_applied': filter_applied,
+    })
+
+
+def stats(request):
+    current_week = datetime.datetime.today().isocalendar()[1]
+    all_submissions = User.objects.all()
+    total_unique_users = len(set(all_submissions.values_list('uuid', flat=True)))
+    current_week_unique = len(set(all_submissions.filter(updated_at__week=current_week).values_list('uuid', flat=True)))
```
umeshksingla: Can you try with `.count()` instead of using `len`?  We only need the count and not the values in memory anyway but I'm not sure if `.count()` works on whatever is returned by `values_list`.

If `values_list` repeats elements and you need unique, try looking for something like `.distinct()`. I'm sure these should be supported as these type of queries are pretty common.

Also, I found out you can actually view the SQL query Django executes which may help you debug in some cases. Example: 
`q = User.objects.filter( ... )`
`print(q.query)`

ports/views.py
```diff
@@ -0,0 +1,293 @@
+from django.shortcuts import render
+from django.http import HttpResponse
+from django.views.decorators.csrf import csrf_exempt
+from django.core.exceptions import ObjectDoesNotExist
+from .models import Port, Category, BuildHistory, Maintainer, Dependency, Builder, User, Variant
+from bs4 import BeautifulSoup
+from django.core.paginator import Paginator, PageNotAnInteger, EmptyPage
+import requests, json, datetime
```
umeshksingla: just to keep all the comments at one place, I'm pasting the link here again: https://www.python.org/dev/peps/pep-0008/#imports

ports/views.py
arjunsalyan: > Can you try with `.count()` instead of using `len`? We only need the count and not the values in memory anyway but I'm not sure if `.count()` works on whatever is returned by `values_list`.

Yes, we certainly can avoid going for sets, thank you. `.distinct()` should be the way to go.

> Also, I found out you can actually view the SQL query Django executes which may help you debug in some cases. Example:
> `q = User.objects.filter( ... )`
> `print(q.query)`

Thank you.

ports/models.py
```diff
@@ -62,7 +62,7 @@ class Builder(models.Model):
 class BuildHistory(models.Model):
     builder_name = models.ForeignKey(Builder, on_delete=models.CASCADE, db_index=True)
     build_id = models.IntegerField()
-    status = models.CharField(max_length=50)
+    status = models.CharField(max_length=50, db_index=True)
     port_name = models.CharField(max_length=50,db_index=True)
     time_start = models.CharField(max_length=50)
```
mojca: You have two serious database issues here.
* You are ordering builds based on `time_start`, but you lack the index on the field.
* You should store this as a timestamp rather than as a char field.

static/js/ajax-all-builds.js
```diff
@@ -0,0 +1,62 @@
...
+function changePage(page) {
+    console.log("Got you");
+    $.ajax({
+        type: 'POST',
+        url: '/all_builds/filter/',
+        data: {
+            'builder': $('#builder-filter').val(),
+            'status': $('#status-filter').val(),
+            'page': page,
+            'csrfmiddlewaretoken': $("input[name=csrfmiddlewaretoken]").val()
```
mojca: What does this do?

static/js/ajax-all-builds.js
```diff
@@ -0,0 +1,62 @@
...
+            beforeSend: function () {
+                $('#all_builds_table').html("Loading Builds...")
+            },
+            dataType: 'html'
+        });
+    });
+});
+
+function changePage(page) {
+    console.log("Got you");
```
mojca: ?

static/js/ajax-all-builds.js
```diff
@@ -0,0 +1,62 @@
...
+            'page': page,
+            'csrfmiddlewaretoken': $("input[name=csrfmiddlewaretoken]").val()
+        },
+        success: filterSuccess,
+        beforeSend: function () {
+            $('#all_builds_table').html("Changing page...")
+        },
+        dataType: 'html'
+    });
+};
```
mojca, line 40: There is a lot of code repetition here, but this part is not really my strong field, so I'm not yet sure what to suggest.

ports/views.py
```diff
@@ -98,34 +99,45 @@ def portdetail(request, name):
...
+        page = request.POST['page']
+
+        if builder == "All Builders" and status == "All Builds":
+            builds = BuildHistory.objects.all().order_by('-time_start')
+        elif builder == "All Builders" and status != "All Builds":
+            builds = BuildHistory.objects.filter(status=status).order_by('-time_start')
+        elif builder != "All Builders" and status == "All Builds":
+            builds = BuildHistory.objects.filter(builder_name__name=builder).order_by('-time_start')
+        else:
+            builds = BuildHistory.objects.filter(status=status, builder_name__name=builder).order_by('-time_start')
```
mojca: I'm not super happy about how this is done. Imagine that you need to apply 10 different filters. Will you really write 1024 if-else sentences? I'm not 100% sure about the correct syntax in Django to achieve that, but this could probably be done with a single `BuildHistory.objects.filter(...)` call with some generalization of the filter calls.

static/js/ajax-all-builds.js
arjunsalyan: I have used a function now.

ports/models.py
arjunsalyan: Earlier I was storing the unix time in database and converting it in the views.
Now, changed to storing the normal time.

> * You are ordering builds based on `time_start`, but you lack the index on the field.

Done.

> * You should store this as a timestamp rather than as a char field.

Done.

ports/views.py
arjunsalyan: Thank you. I have now switched to using django-filters and this is no longer needed. To maintain consistency, I will switch all the filters to use django-filters.

static/js/ajax-all-builds.js
arjunsalyan: Removed. Was trying to catch an error with pagination, forgot to remove it.

static/js/ajax-all-builds.js
arjunsalyan: This needs to be passed with all POST requests- built in support for additional security.

mojca:
I'm merging this PR. I will run another batch of reviews later, but if we keep postponing the merge, getting further code reviews done will be increasingly more difficult. I'm happy that we got rid of giant files, other things that might be still missing can be addressed in separate tickets.

PR #6, opened 2019-06-07
Add .travis.yml

arjunsalyan:


mojca:
This looks almost good, we just need to get rid of the build failure, and probably update the version of postgresql.

arjunsalyan:
> This looks almost good, we just need to get rid of the build failure, and probably update the version of postgresql.

Turns out there is no python 3.7 for trusty yet. We will have to use Xenial instead. I am not sure if we should shift all builds to Xenial or just the python 3.7 one.

UPDATE: So, I currently changed it to use Xenial only for python 3.7, and all builds are now passed.

mojca:
We could easily go for `disco`, `bionic` and a newer version of postgres.

arjunsalyan:
> We could easily go for `disco`, `bionic`

[https://docs.travis-ci.com/user/reference/overview/#linux](https://docs.travis-ci.com/user/reference/overview/#linux) ,  I am not sure of this- but maybe these are the only choices.

arjunsalyan:
postgres 10 fails on 14.04. Maybe we should use 16.04 for all builds?

mojca:
May I request squashing the first four commits and remove the last two from this particular pr? They are fine, but don't belong to Travis. You can cherry-pick them to a different branch or ask for help with git magic if you need it.

arjunsalyan:
> May I request squashing the first four commits and remove the last two from this particular pr? They are fine, but don't belong to Travis. You can cherry-pick them to a different branch or ask for help with git magic if you need it.

Okay, thanks.

arjunsalyan:
Done

mojca:
I'm merging this as is, as it's good enough for now.

My general comment is that we don't need to worry about backward compatibility with ancient systems (as we might need to do if we had software that was supposed to run on multiple platforms). So I don't really care if we don't test on older versions of python/django/postgres/django. We could easily test just on Ubuntu versions `disco` and `bionic`, with just python 3.7, with just django 2.2.2, with just the latest postgresql.

What would be the added benefit would be checking if this works on macports, but this is a bit trickier to do as you would first need to install macports and all the dependencies. Something for another time, maybe.

Another minor thing is that coverage report can certainly be done with a separate result being shown, like here for example: https://github.com/buildbot/buildbot/blob/master/README.rst

arjunsalyan:
> What would be the added benefit would be checking if this works on macports, but this is a bit trickier to do as you would first need to install macports and all the dependencies. Something for another time, maybe.

Sorry, but I am not exactly sure what this means. Does this mean bundling it as a port?

> Another minor thing is that coverage report can certainly be done with a separate result being shown, like here for example: https://github.com/buildbot/buildbot/blob/master/README.rst

This can be achieved with coveralls.io or codecov.io- I tried but I am unable to get access to MacPorts' repositories through them.

mojca:
> > What would be the added benefit would be checking if this works on macports, but this is a bit trickier to do as you would first need to install macports and all the dependencies. Something for another time, maybe.
> 
> Sorry, but I am not exactly sure what this means. Does this mean bundling it as a port?

No. At the moment you only run the test on linux. You could run it on Mac. You don't need to bundle anything into a port, but if you use macOS, Travis comes preconfigured with brew, so you can easily install something via brew. On the other hand if you want to install dependencies needed to run the tests via macports, you need to go through quite a bit of bootstrapping, which makes the whole process a bit less straightforward. (Too much work at this very moment.)

> > Another minor thing is that coverage report can certainly be done with a separate result being shown, like here for example: https://github.com/buildbot/buildbot/blob/master/README.rst
> 
> This can be achieved with coveralls.io or codecov.io- I tried but I am unable to get access to MacPorts' repositories through them.

Oh, do we need to enable this explicitly?

umeshksingla:
It would need to be enabled explicitly by the owner of the repository, yes. Then, we can use `codecov` to use our `tox` configuration for running tests in a Travis step. I'll send the request once we decide on which tool. If none of us are experienced with any of the tools, let's just use `codecov` since buildbot uses it. The repository owner simply needs to give access to `codecov` as a third party plugin, as is with most tools like Travis, Gitter, Jenkins, etc.

Issue #7, opened 2019-06-08
Tabbed views on Port-detail


arjunsalyan:

mojca:
Could we add at least some anchor to select a particular view? So that we could, say, link directly to the build status or statistics of a port? At the moment one can only link to the summary page.

Another minor request: can we add back the numbers of trac tickets? I would simply put the number at the front (on the left) and make it a clickable link to the trac ticket. With that in place maybe "View on Trac" can even go away.

mojca:
I would also suggest printing the name and description of the port above the tabs and always have it visible, also when viewing the build or installation statistics, tickets etc. And maybe put `View on GitHub` to the right of port name?

Probably port health could also be above statistics.

![image](https://user-images.githubusercontent.com/15606/59549157-cc106a80-8f59-11e9-8e01-d43449776311.png)

arjunsalyan:
> Another minor request: can we add back the numbers of trac tickets? I would simply put the number at the front (on the left) and make it a clickable link to the trac ticket. With that in place maybe "View on Trac" can even go away.

I will see if this is possible, thanks.

PR #13, opened 2019-06-09
Workspace

arjunsalyan:
This is the area where all the commits comes in first. Most of the code feedback happens here (Week 3 and onwards).

- [x] Remove glitches in build history and combine all failed builds together
- [x] Tabbed sections on Port-detail
- [x] Optimisation of All Builds
- [ ] Optimisation of Updating and Loading Scripts.
- [ ] Increase Test Coverage
- [ ] Take care of reasonable lint warnings


ports/models.py
```diff
@@ -63,7 +63,7 @@ class BuildHistory(models.Model):
     builder_name = models.ForeignKey(Builder, on_delete=models.CASCADE, db_index=True)
     build_id = models.IntegerField()
     status = models.CharField(max_length=50, db_index=True)
-    port_name = models.CharField(max_length=50,db_index=True)
+    port_name = models.CharField(max_length=50, db_index=True)
```
mojca: Generally we prefer to do formatting changes in separate commits.

Let's say that you need to revert a particular change, or debug something with bisection. In such cases it's much easier if changes are separate.

(Tox should automatically catch these cases if configured properly.)

ports/templates/ports/all_builds.html
```diff
@@ -20,9 +20,8 @@ <h4>Builds</h4>
             <label for="status-filter">Filter based on Status: </label>
             <select class="form-control ml-2 filter" id="status-filter" name="status-filter">
                 <option value="">All</option>
-                <option value="build successful">build successful</option>
-                <option value="failed install-port">failed install-port</option>
-                <option value="failed install-dependencies">failed install-dependencies</option>
+                <option value="build successful">Successful Builds</option>
+                <option value="failed">Failed Builds</option>
```
mojca: After something thinking: what about keeping both of these new ones (Successful Builds & Failed Builds), but adding the third one that's equivalent to the previous "failed install-port"? I'm not sure about the best names, but sometimes one would probably want to ignore the failures with broken dependencies.

ports/views.py
```diff
@@ -1,16 +1,16 @@
...
+import requests
 from bs4 import BeautifulSoup
 from django.shortcuts import render
 from django.http import HttpResponse
 from django.views.decorators.csrf import csrf_exempt
 from django.core.paginator import Paginator, PageNotAnInteger, EmptyPage
 
-from .models import Port, Category, BuildHistory, Maintainer, Dependency, Builder, User, Variant
 from parsing_scripts import update
+from .models import Port, Category, BuildHistory, Maintainer, Dependency, Builder, User, Variant
```
mojca: Again, pure formatting changes probably make more sense as separate commits.

mojca (commented):
This generally looks good to me, just some nitpicking:
* I would use separate commits for formatting changes.
* Try to make the commits as small as possible. Like: the `time_build` fix can go into its own commit. `filters.py` probably need the change in `all_builds.html` to work correctly. Changing `id` to `port_id` can go to its own commit, the `else return HttpResponse ...` can go to its own commit ...

No need to open ten different pull requests for that, just split the commits in the smallest logical units. Once you do this, I believe this is good to go.

arjunsalyan:
> This generally looks good to me, just some nitpicking:
> 
> * I would use separate commits for formatting changes.
> * Try to make the commits as small as possible. Like: the `time_build` fix can go into its own commit. `filters.py` probably need the change in `all_builds.html` to work correctly. Changing `id` to `port_id` can go to its own commit, the `else return HttpResponse ...` can go to its own commit ...
> 
> No need to open ten different pull requests for that, just split the commits in the smallest logical units. Once you do this, I believe this is good to go.

Thank You. I have tried to separate out commits and will keep this in mind in future. Please let me know if something still needs to be taken care of.

arjunsalyan:
I will be using this pull request for all commits of this week, if that is not a problem?

But now I feel that having a separate pull request for the OS-chart would be a better idea.

mojca:
You may now rebase the commits in this PR.

ports/templates/ports/stats.html
```diff
@@ -15,17 +54,31 @@
...
+    <hr>
+<div class="row">
+    <div class="col-lg-6 text-center">
+        <div id="chart_div"></div>
+    </div>
+    <div class="col-lg-6 text-center">
+        <br>
+        <h5>What does this chart show?</h5>
+        <p>This chart shows the distribution of <b>OSX Version</b> vs the <b>number of unique users</b> for the month
+            ongoing month.
```
mojca: "For the month ongoing month" sounds a bit strange.

ports/templates/ports/builds_filtered_table.html
```diff
@@ -54,12 +54,12 @@ <h4 class="text-secondary">Showing {% if port_name != '' %}results for <i>"{{ po
     </thead>
     {% for build in builds %}
         <tr>
-            <td><a href="/ports/{{ build.port_name }}">{{ build.port_name }}</a></td>
+            <td><a href="{% url 'port_detail' build.port_name %}">{{ build.port_name }}</a></td>
             <td>{{ build.builder_name.name }}</td>
-            <td><a target="_blank" href="{{ build.build_url }}">{{ build.build_id }}</a></td>
-            <td>{{ build.time_start }} </td>
+            <td><a target="_blank" href="https://build.macports.org/builders/ports-{{ build.builder_name.name }}-builder/builds/{{ build.build_id }}">{{ build.build_id }}</a></td>
```
mojca: I would turn this into a simple python function and call the function from the template instead.

ports/templates/ports/builds_filtered_table.html
```diff
@@ -54,12 +54,12 @@ <h4 class="text-secondary">Showing {% if port_name != '' %}results for <i>"{{ po
...
+            <td>{{ build.time_start|date:"Y-m-d" }}  {{ build.time_start|time:"G:i:s" }} </td>
             <td>{{ build.time_elapsed|time:"H:i:s" }}</td>
-            <td><a target="_blank" href="{{ build.watcher_url }}">{{ build.watcher_id }}</a></td>
+            <td><a target="_blank" href="https://build.macports.org/builders/ports-{{ build.builder_name.name }}-watcher/builds/{{ build.watcher_id }}">{{ build.watcher_id }}</a></td>
```
mojca: Same as above. I would be tempted to say that both could be displayed in the same column, but then ... I mostly don't want to click on the watcher, I usually want to click on the builder, and it needs to be clear which one is which. Most users don't care about the build number, and would just occasionally want to click on the link. So the text for the link could even be much shorter. Still looking for ideas.

ports/templates/ports/builds_filtered_table.html
```diff
@@ -54,12 +54,12 @@ <h4 class="text-secondary">Showing {% if port_name != '' %}results for <i>"{{ po
     </thead>
     {% for build in builds %}
         <tr>
-            <td><a href="/ports/{{ build.port_name }}">{{ build.port_name }}</a></td>
+            <td><a href="{% url 'port_detail' build.port_name %}">{{ build.port_name }}</a></td>
             <td>{{ build.builder_name.name }}</td>
```
mojca: It would make sense to also call a python function here that would turn `builder_name.name` into a shorter version. For anything beyond 10.6 this could always be just OS number.

ports/templates/ports/port-detail/build_information.html
```diff
@@ -0,0 +1,94 @@
...
+                    {% if builds.has_next %}
+                        <li class="page-item"><button class="page-link" onclick="changePage({{ builds.next_page_number }})">&raquo;</button>
+                        </li>
+                    {% else %}
+                        <li class="disabled page-item"><span class="page-link">&raquo;</span></li>
+                    {% endif %}
+                </ul>
+            </nav>
+{% endif %}
+<table class="table table-striped">
```
mojca, line 69: Just curious: would this still work / would it make sense if you separate this table into its own template and include it here? Same question for other tables. Again: I have no strong preference, just brainstorming.

ports/views.py
```diff
@@ -277,6 +330,19 @@ def search_ports_in_maintainer(request):
 
 
 # Accept submissions from mpstats and update the users table
+def populate_os_table(user_obj):
```
mojca: I would like to solve this in a different way, but that's a topic to be discussed separately. At the moment it's really awesome to see some actual statistics appear on the site.

ports/views.py
```diff
@@ -78,32 +78,79 @@ def variantlist(request, variant):
...
+            result = paginated_builds.get_page(paginated_builds.num_pages)
+
+        return render(request, 'ports/port-detail/build_information.html', {
+            'builds': result,
+            'builder': builder,
+            'builders': builders,
+            'status': status,
+        })
+    except Port.DoesNotExist:
+        return HttpResponse("<br><br><h3>Sorry, the port you requested was not found.</h3><br>")
```
mojca: No super strong feelings, but as soon as you need that much formatting, it might be time for a template?

Anyway, I'm much happier now than I was with the previous response.

arjunsalyan: Yes, I would also try to cover all such exceptions across the app.

ports/templates/ports/port-detail/build_information.html
arjunsalyan: Yes, it would work but we would end up increasing the layers in the nested requests. This is not very recommended because it makes maintenance difficult. 

>Same question for other tables.

We at most places are using a separate table loaded through an Ajax-request (search. It is also almost the case here- just that we have filters and pagination also on the same page. I am not sure if we should go for a third layer.

ports/templates/ports/builds_filtered_table.html
arjunsalyan: Thank you, I have done it. Just need to take care of 10.5.

ports/templates/ports/builds_filtered_table.html
arjunsalyan: Thank you

ports/templates/ports/stats.html
arjunsalyan: Oops!

ports/templates/ports/builds_filtered_table.html
arjunsalyan: It is showing some strange behaviour with 10.14, will fix this

ports/templates/ports/port-detail/summary.html
```diff
@@ -0,0 +1,114 @@
+{% load format_names %}
+
+<h3>{{ port.portdir }} | {{ port.name }}</h3>
```
umeshksingla: I just have a little feedback. In the tabbed view, when I switch from the summary tab to some other tab (builds or tickets or installs) I do not see for what port the information is being shown. IMO it'd be better to have the port name above those 4 tabs. Example:

Screen Shot 2019-06-15 at 12 36 09 AM

arjunsalyan: Thank you, I have done the changes: also mentioned here: https://github.com/macports/macports-webapp/issues/7#issuecomment-502347279

mojca: Yes, this was done and I'm happy with the result.

.travis.yml
```diff
@@ -17,7 +17,8 @@ env:
     - SECRET_KEY=thISkEyisUsedForTravis
 # install dependencies
 install:
-  - pip install -r requirements.txt
+  - pip install -q Django==$DJANGO_VERSION
+  - pip install -q -r requirements.txt
```
umeshksingla: There's `Django==2.1.7` in requirements.txt which probably needs to be removed.

parsing_scripts/git_update.py
```diff
@@ -0,0 +1,91 @@
...
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+sys.path.append(BASE_DIR)
+os.environ['DJANGO_SETTINGS_MODULE'] = 'MacPorts.settings'
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "MacPorts.settings")
+
+django.setup()
+
+
+def clone_repo():
+    if os.path.isdir('macports-ports'):
```
umeshksingla: Shouldn't we check for `.git` directory inside `macports-ports` instead?

parsing_scripts/git_update.py
```diff
@@ -0,0 +1,91 @@
...
+            if remote == "https://github.com/macports/macports-ports.git":
+                print("The macports-ports repo is healthy.")
+            else:
+                raise OSError
+
+            # If old_hash is not provided by the user
+            if old_hash is False:
+                try:
+                    # Try to fetch the most recent hash from database
+                    # Try to fetch the most recent hash from database
```
umeshksingla: repeated comment as above

parsing_scripts/git_update.py
```diff
@@ -0,0 +1,91 @@
...
+            os.chdir(BASE_DIR)
+            return updated_ports
+
+        except OSError:
+            os.chdir(root)
+            print("macports-ports repository has some error")
+            print("Cleaning current tree and cloning new repo.")
+            shutil.rmtree('macports-ports')
+            clone_repo()
+            return get_list_of_changed_ports(new_hash, old_hash, root)
```
umeshksingla, line 89: If there's an `OSError` which keeps repeating and needs a manual resolution, the script will just run infinitely. Needs an improvement.

parsing_scripts/git_update.py
```diff
@@ -0,0 +1,91 @@
...
+
+            # Pull from macports-ports
+            subprocess.call(['git', 'pull'])
+
+            # Get new hash if not provided
+            if new_hash is False:
+                new_hash = subprocess.run(['git', 'rev-parse', 'HEAD'], stdout=subprocess.PIPE).stdout.decode('utf-8')
+
+            # find the range of commits to loop over
+            range = str(old_hash).strip() + "^.." + str(new_hash).strip()
```
umeshksingla, line 67: "to loop over" is probably not the best way to describe what's happening

parsing_scripts/parse_build_history.py
```diff
@@ -109,7 +109,7 @@ def fetch():
         if build_number_loaded:
             build_in_database = build_number_loaded[0].build_id + 1
         else:
-            build_in_database = last_build_number - 200
+            build_in_database = last_build_number - 9999
 
         # lets start with the latest four builds
```
umeshksingla: "four"?

ports/filters.py
```diff
@@ -12,6 +12,14 @@ class Meta:
         fields = ['builder_name__name']
 
 
+class BuildHistoryFilterForPort(django_filters.FilterSet):
```
umeshksingla: I'd need a little help in understanding this. How is this different from the BuildHistoryFilter just above?

Also, I do not see the need to do a `contains` query in cases where the set of possible values for a field is fixed, e.g., status, builders in this case. An `exact` query should be performed instead.

ports/models.py
```diff
@@ -80,3 +78,21 @@ class User(models.Model):
     os_arch = models.CharField(max_length=10)
     full_json = JSONField()
     updated_at = models.DateTimeField(auto_now=True)
+
+    def get_year(self):
```
umeshksingla: `@property` decorators are used for similar purposes. 
```
@property
def get_year(self):
    return self.updated_at.year
```
Then you can access the attribute by `User.get_year` instead of `User.get_year()`. You don't even need to write get_ then, just simply define `def year(self):` and access by `User.year`. 

None of these help you with queries happening at the database level though. A short read [here](http://dylanbfox.blogspot.com/2015/01/django-tip-using-property-decorator-on.html).

ports/templates/400.html
```diff
@@ -0,0 +1,8 @@
+{% extends 'ports/layout.html' %}
+{% block content %}
+<br>
+<br>
+<div class="text-center container" style="max-width: 700px">
+    <h3>400 Bad Request</h3>
+</div>
+{% endblock %}
```
umeshksingla: A newline at the end of file, please.

ports/templates/404.html
```diff
@@ -0,0 +1,10 @@
+{% extends 'ports/layout.html' %}
+{% block content %}
+<br>
+<br>
+<div class="text-center container" style="max-width: 700px">
+    <h3>404 Page Not Found</h3>
+    <p>We encountered a 404 error when trying render the requested resource. If you think this is a mistake,
+    please report.</p>
+</div>
+{% endblock %}
```
umeshksingla: \n

ports/templates/500.html
```diff
@@ -0,0 +1,10 @@
+{% extends 'ports/layout.html' %}
+{% block content %}
+<br>
+<br>
+<div class="text-center container" style="max-width: 700px">
+    <h3>500 Internal Server Error</h3>
+    <p>Please try again. If the problem persists, kindly report.
+    </p>
+</div>
+{% endblock %}
```
umeshksingla: \n

ports/templatetags/url_generate.py
```diff
@@ -0,0 +1,12 @@
+from django import template
+register = template.Library()
+
+
+@register.simple_tag
+def build_url(builder_name, build_id):
+    return "https://build.macports.org/builders/ports-" + builder_name + "-builder/builds/" + str(build_id)
```
umeshksingla: would be better if the base URL ("https://build.macports.org/builders/ports-") is not embedded here.

ports/templatetags/format_names.py
```diff
@@ -0,0 +1,23 @@
+from django import template
+
+register = template.Library()
+
+
+@register.filter('os_version')
+def trim_builder_name(builder_name):
+    first_trim = builder_name.replace('_x86_64', '')
+    return first_trim.replace('_legacy', '')
```
umeshksingla: can simply be `return builder_name.replace('_x86_64', '').replace('_legacy', '')`

ports/templates/ports/exceptions/category_not_found.html
```diff
@@ -0,0 +1,10 @@
+{% extends 'ports/layout.html' %}
+
+{% block content %}
+<br>
+<br>
+<div class="text-center container" style="max-width: 700px">
+    <h3>Are you looking for the right category?</h3>
+    <p>We could not find that category. If you think this is a mistake, please report.</p>
+</div>
+{% endblock %}
```
umeshksingla: \n

ports/templates/ports/exceptions/port_not_found.html
```diff
@@ -0,0 +1,9 @@
+{% extends 'ports/layout.html' %}
+
+{% block content %}
+<br><br>
+<div class="text-center">
+    <h3>Sorry, we could not find any port for "{{ name }}".</h3>
+    <p>If you think a port is missing, please report.</p>
+</div>
+{% endblock %}
```
umeshksingla: \n

ports/models.py
```diff
@@ -80,3 +78,21 @@ class User(models.Model):
...
+        return self.updated_at.year
+
+    def get_month(self):
+        return self.updated_at.month
+
+
+class OSDistribution(models.Model):
```
umeshksingla, line 432: I think I missed the discussion but why exactly is this model class needed? @mojca I think we should move that database readme from last year's repo to this one for easy access.

ports/templates/ports/port-detail/summary.html
```diff
@@ -0,0 +1,89 @@
...
+                    <th scope="row">Homepage</th>
+                    <td><a href="{{ port.homepage }}" target="_blank">{{ port.homepage }}</a></td>
+                </tr>
+                <tr>
+                    <th scope="row">Version:</th>
+                    <td>{{ port.version }}</td>
+                </tr>
+                <tr>
+                    <th scope="row">Maintained By:</th>
+                    <td>{% if maintainers.all|length > 0 %}
```
umeshksingla: probably move the `if` condition to below `<td>` to keep the if-else-endif in the same line.

ports/templates/ports/port-detail/summary.html
```diff
@@ -0,0 +1,89 @@
...
+                    <th scope="row">License:</th>
+                    <td>{{ port.license }}</td>
+                </tr>
+                <tr>
+                    <th scope="row">Platforms:</th>
+                    <td>{{ port.platforms }}</td>
+                </tr>
+                <tr>
+                    <th scope="row">Categories:</th>
+                    <td>{% for cat in port.categories.all %}
```
umeshksingla: same here. would look better if `{% for ...` starts in the next line after `<td>`.

umeshksingla (commented):
I was thinking if it is possible to move all the preprocessing code from `parse_build_history.py` script to the model class so when there's a change in the model, the "nearby" code reminds one to look into the populate method as well. The script should be kept minimal in this case:
```
import ...
from ports.models import BuildHistory

BuildHistory.populate()
```
This script can then be scheduled and the need to modify code outside the project will be very less. At the moment, this part feels disconnected from the other code whereas it is equally important to be maintained.

Same for the case while updating portindex.

ports/filters.py
arjunsalyan: > I'd need a little help in understanding this. How is this different from the BuildHistoryFilter just above?

`BuildHistoryFilterForPort` is definitely redundant. Thank you

> Also, I do not see the need to do a `contains` query in cases where the set of possible values for a field is fixed, e.g., status, builders in this case. An `exact` query should be performed instead.

We are also running a filter for "All Failed Builds" which includes all failed builds regardless of `failed install-port` or `failed install-dependencies`. So I just pass 'failed' and contains helps with that.

ports/models.py
arjunsalyan: We haven't discusses about it yet. I just followed the plan from my proposal- where the idea is to have separate tables for ports, os etc.

ports/templatetags/url_generate.py
arjunsalyan: I am not sure if I get your point, what else do you suggest?

ports/models.py
arjunsalyan: Thank you so much, I just realised that I am not even using this method, so we can get rid of it.

parsing_scripts/git_update.py
arjunsalyan: We can check for `.git`. But I am first checking if the `macports-ports` directory exists at all.

Even if we check for `.git`, we will have to check for the right `remote`. But checking for right `remote` removes the need for checking `.git`. Is there something I am missing?

parsing_scripts/git_update.py
arjunsalyan: There are two lines which can raise an `OSError`:

1. If the remote is wrong
2. If there is a problem with `git pull`

And as soon as an `OSError` is raised, a fresh copy of macports-ports is cloned. I thought that a new copy should solve both of the above problems. But maybe we need a way to break the recursion at second run.

arjunsalyan: There are two lines which can raise an `OSError`:

1. If the remote is wrong
2. If there is a problem with `git pull`

And as soon as an `OSError` is raised, a fresh copy of macports-ports is cloned. I thought that a new copy should solve both of the above problems. But maybe we need a way to break the recursion at second run.

ports/templates/400.html
arjunsalyan: Thank you.

parsing_scripts/git_update.py
arjunsalyan: Thank You

arjunsalyan:
> I was thinking if it is possible to move all the preprocessing code from `parse_build_history.py` script to the model class so when there's a change in the model, the "nearby" code reminds one to look into the populate method as well. The script should be kept minimal in this case:
> 
> ```
> import ...
> from ports.models import BuildHistory
> 
> BuildHistory.populate()
> ```
> 
> This script can then be scheduled and the need to modify code outside the project will be very less. At the moment, this part feels disconnected from the other code whereas it is equally important to be maintained.
> 
> Same for the case while updating portindex.

This seems like a good idea. I have already created custom management commands using these scripts:
```
python manage.py load {filename}
python manage.py update {filename}
python manage.py fetch-build-history
```
I will work on this.

arjunsalyan:
@umeshksingla: Please see this: 51624d1 . Did I get your point right?

arjunsalyan:
> IMO you don't need a class here. A class method will suit here well. You can use inner functions here if you do not want to expose all the helper methods but only `populate()`. A short read [here](https://stackabuse.com/python-nested-functions/).

Thank you. I have tried to do this in d27b625.

ports/views.py
```diff
@@ -67,13 +67,17 @@ def variantlist(request, variant):
 # Views for port-detail page START
 def portdetail(request, name):
     try:
+        req_port = Port.objects.get(name__iexact=name)
+
+        all_latest_builds = BuildHistory.objects.all().order_by('port_name', 'builder_name', '-build_id').distinct('port_name', 'builder_name')
```
mojca, line 78: I didn't investigate yet, but the builds are no longer displayed correctly. If you look at gnuplot, most of the builds are missing in the port health status.

MacPorts/settings.py
```diff
@@ -147,6 +147,7 @@
 # Configuration for CORS
 CORS_ORIGIN_WHITELIST = [
     'http://localhost:8010',
+    'http://34.67.147.76',
```
mojca, line 149: In a way I would prefer if this commit only went to the branch that gets deployed (and if that special branch would be pushed to this repository). The settings are just temporary.

ports/views.py
arjunsalyan: Yes, I am aware of the issue. This is a very peculiar problem. I have tried two ways I could think of doing this and both gave same incorrect results.

The version deployed on EC2 does not contain these commits and I see the exact same results there also. However, it was working fine  last I checked.

arjunsalyan: Yes, I am aware of the issue. This is a very peculiar problem. I have tried two ways I could think of doing this and both gave same incorrect results.

The version deployed on EC2 does not contain these commits and I see the exact same results there also. However, it was working fine  last I checked.

**UPDATE 1** : This problem is strange and so far I am unable to find what is going wrong. http://ec2-52-34-234-111.us-west-2.compute.amazonaws.com/port/pgbadger/ if we see this, the behaviour is perfect, and then for some port it suddenly behaves strange. The results are same for both new method (on heroku) and the old one (on ec2).

The old method should not go wrong:
```
latest_builds[builder] = BuildHistory.objects.filter(builder_name__name=builder, port_name=name).order_by('-time_start').first()

```
I don't understand how can this go wrong.
I will write tests for build history.

**UPDATE 2**: (fixed) The problem was not with `port health` but with the `build information` being displayed on the `port-detail` page. The lookup expression was `icontains` and this made history of other ports to be displayed on the page as well. If we now go to [gnuplot](https://frozen-falls-98471.herokuapp.com/port/gnuplot/), the build info is only available for two builders. And Port health is also displayed correctly.

MacPorts/settings.py
arjunsalyan: Okay, thank you. Currently, this branch (week3/improvements) gets deployed.

To clarify, the special branch you are talking about should also be pushed here?

ports/views.py
```diff
@@ -205,92 +259,89 @@ def maintainer_detail_email(request, name, domain):
...
-
-            user = User()
-            user.uuid = received_json['id']
-            user.osx_version = received_json['os']['osx_version']
-            user.macports_version = received_json['os']['macports_version']
-            user.os_arch = received_json['os']['os_arch']
-            user.xcode_version = received_json['os']['xcode_version']
-            user.full_json = received_json
-            user.save()
+            submission_id = Submission.populate(received_json)
```
umeshksingla: You can get the `submission_timestamp` here using `datetime.datetime.utcnow()` and store it in the submissions table. Look for the appropriate format to store in Postgres.

.travis.yml
umeshksingla: I think this probably got overlooked.

api_v1/serialisers.py
```diff
@@ -0,0 +1,49 @@
+from rest_framework import serializers
+
+from ports.models import Port, Maintainer, Variant, BuildHistory, Builder
+
+
+class MaintainerSerialiser(serializers.ModelSerializer):
```
umeshksingla: I don't know if it matters but the serializers are spelled with 'z' in the `rest_framework`. Should we change `MaintainerSerialiser ` to `MaintainerSerializer ` to keep it consistent? Same for other class names in the module.

ports/models.py
```diff
@@ -62,13 +297,113 @@ class Builder(models.Model):
 class BuildHistory(models.Model):
     builder_name = models.ForeignKey(Builder, on_delete=models.CASCADE, db_index=True)
```
umeshksingla: Will it be better to keep all the indexes at one place, in class `Meta` for example?

ports/models.py
```diff
@@ -62,13 +297,113 @@ class Builder(models.Model):
...
+                data = json.loads(u.read().decode())
+            return data
+
+        def get_build_properties(array):
+            properties = {}
+            for prop in array['properties']:
+                properties[prop[0]] = prop[1]
+            return properties
+
+        def get_build_steps(array):
```
umeshksingla: I think this is unused.

parsing_scripts/git_update.py
umeshksingla: I am worried in case `OSError` gets thrown in some case other than the two you mentioned above.

parsing_scripts/git_update.py
umeshksingla: How about checking for `macports-ports/.git`? I just got this idea from `mprsyncup` in macports-infrastructure...

parsing_scripts/git_update.py
```diff
@@ -0,0 +1,90 @@
...
+    if os.path.isdir('macports-ports'):
+        return
+    else:
+        os.system("git clone https://github.com/macports/macports-ports.git")
+
+
+def get_list_of_changed_ports(new_hash=False, old_hash=False, root=BASE_DIR):
+    os.chdir(root)
+
+    if os.path.isdir('macports-ports'):
```
umeshksingla: Also, it'll be better to take `macports-ports` in a variable than to repeat the repo name at multiple places.

ports/management/commands/fetch-build-history.py
```diff
@@ -1,11 +1,11 @@
 from django.core.management.base import BaseCommand, CommandError
 
-from parsing_scripts import parse_build_history
```
umeshksingla, line 3: I think `parse_build_history.py` can be removed now unless you have kept it for some reasons?

ports/management/commands/load.py
```diff
@@ -1,6 +1,6 @@
 from django.core.management.base import BaseCommand, CommandError
 
-from parsing_scripts import load_initial_data
```
umeshksingla, line 3: same for `load_initial_data`.

ports/management/commands/update.py
```diff
@@ -1,6 +1,6 @@
 from django.core.management.base import BaseCommand, CommandError
 
-from parsing_scripts import update
```
umeshksingla, line 3: same for `update`.

ports/models.py
```diff
@@ -28,6 +33,236 @@ class Port(models.Model):
...
+            Category.objects.bulk_create(batch)
+            return
+
+        def load_ports_and_maintainers_table(ports):
+            for port in ports:
+
+                # Add Ports to the Database One-by-One
+                new_port = Port()
+                try:
+                    new_port.name = port['name']
```
umeshksingla, line 63: So, does this mean `name` attribute is not necessary? Same for `portdir` and `version`.

I may have missed conversation around this but if there is something, it'd help if there's a comment on top since this is not trivial IMO. Weren't we having conversations around making the `name` as the key?

Won't this need to specify `null=True` or equivalent in the `Port` model above if that is the case indeed?

ports/models.py
```diff
@@ -28,6 +33,236 @@ class Port(models.Model):
 
     objects = PortManager()
 
+    @classmethod
+    def load(cls, data):
+        def open_portindex_json(path):
+            with open(path, "r") as file:
```
umeshksingla, line 40: Let's just catch the error in case the `path` does not exist and notify the user instead of iterating through all the functions and doing nothing.

ports/models.py
```diff
@@ -28,6 +33,236 @@ class Port(models.Model):
...
+
+                except Port.DoesNotExist:
+                    print("Failed to update dependencies for {}. Port not found in database.".format(port['name']))
+        if isinstance(data, list):
+            load_ports_and_maintainers_table(data)
+            load_categories_table(data)
+            load_dependencies_table(data)
+        elif isinstance(data, str):
+            ports = open_portindex_json(data)
+            load_categories_table(ports)
```
umeshksingla, line 138: Maybe have a small method for these 3 lines separately? We do not want to forget and repeat in case it supports a new format of input in future...

ports/models.py
```diff
@@ -28,6 +33,236 @@ class Port(models.Model):
...
+                            obj.dependencies.add(*dependencies)
+
+                except Port.DoesNotExist:
+                    print(
+                        "Failed to update depencies for {}. Port does not exist in the database.".format(port['name']))
+
+        if isinstance(data, str):
+            if is_json:
+                ports = open_portindex_json(data)
+                full_update_ports(ports)
```
umeshksingla: Same for the two calls here which are repeated 3 times already,

ports/models.py
umeshksingla: Okay.

ports/models.py
```diff
@@ -28,6 +33,236 @@ class Port(models.Model):
...
+            load_categories_table(data)
+            load_dependencies_table(data)
+        elif isinstance(data, str):
+            ports = open_portindex_json(data)
+            load_categories_table(ports)
+            load_ports_and_maintainers_table(ports)
+            load_dependencies_table(ports)
+
+    @classmethod
+    def update(cls, data, is_json=True):
```
umeshksingla, line 151: It seems to me that we are anyway supporting the full `load` functionality in `update`. Do you think we can write `update` slightly more cleverly and remove the need for `load` completely?
cc @mojca

ports/models.py
```diff
@@ -62,13 +297,113 @@ class Builder(models.Model):
...
+            build.build_id = data['buildnr']
+            build.time_start = data['time_start']
+            build.time_elapsed = data['buildtime']
+            build.builder_name = builder
+            build.build_url = data['url']
+            build.watcher_url = data['watcher_url']
+            build.watcher_id = data['watcher_id']
+            build.save()
+
+        fulldata = {}
```
umeshksingla: Maybe move it to just before the loop `for buildername in builders:`.

ports/templatetags/url_generate.py
umeshksingla: Simply having something like `url_prefix = 'https://build.macports.org'` on top.

umeshksingla (changes requested):
I have tried reviewing most of the main code but there's still a lot left. I hesitate to merge it to master today but we are close.

@arjunsalyan Good job for a lot of functionality in such little time. I have left some comments. Feel free to disagree and discuss.

@mojca There's a huge component of HTML code, and there's unit-tests and there's JS code I still haven't touched at all. Do you think we should split the task of reviewing, at least for this PR, component wise? Suggest something.

api_v1/serialisers.py
arjunsalyan: Thank you. I will do this.

ports/models.py
arjunsalyan: Yeah, definitely. Thanks

ports/models.py
arjunsalyan: Maybe I have implemented it in a wrong way, but the idea is to not add the port if any of these fields (name, version, portdir) are missing. That is why I added `continue` in the `except` block. Is something wrong with the code?

There should not be cases where any of these keys is missing. But in case if there is one- we would not want to add the port because something is wrong with it.

ports/models.py
arjunsalyan: `path` is passed on to this function by a management command- which throws an error if the file was not found.

If we need to catch the error here also, I will add it.

**UPDATE**: Okay, it will definitely be better if we catch the error here also. Thank you

ports/models.py
arjunsalyan: Okay, thank you

UPDATE: Done.

ports/models.py
arjunsalyan: Done.

ports/models.py
arjunsalyan: This is needed by the function `add_to_summary`.

umeshksingla: Then I think it would be better to return it from `add_to_summary`.

ports/models.py
umeshksingla: Okay, my bad misinterpreting the `continue` with `pass`, thank you.

ports/models.py
umeshksingla: Now that I think about your comment, you are right that it will simply fail at `open` and it seems like catching the exception here could be an over-the-top.

ports/models.py
umeshksingla: I think you probably forgot to push them or something. I just pulled in the latest code and I don't see the changes.

ports/models.py
umeshksingla: Just use your best judgment here.

ports/templatetags/url_generate.py
```diff
@@ -4,9 +4,11 @@
...
 def build_url(builder_name, build_id):
-    return "https://build.macports.org/builders/ports-" + builder_name + "-builder/builds/" + str(build_id)
+    url_prefix = "https://build.macports.org/builders/ports-"
+    return url_prefix + builder_name + "-builder/builds/" + str(build_id)
 
 
 @register.simple_tag
 def watcher_url(builder_name, watcher_id):
-    return "https://build.macports.org/builders/ports-" + builder_name + "-watcher/builds/" + str(watcher_id)
+    url_prefix = "https://build.macports.org/builders/ports-"
```
umeshksingla: Actually, the entire point of having the base URL in a constant var was to avoid changing the URL in the future at multiple places.

ports/models.py
arjunsalyan: I did the changes in: 22a718e .

Here is the the line from my fork: https://github.com/arjunsalyan/macports-gsoc-2019-webapp/blob/0c5f1491dc886ec530a3c947f6e4ac1839964d8f/ports/models.py#L255

ports/models.py
arjunsalyan: I have added a check while calling the function `open_portindex_json`.

ports/filters.py
umeshksingla: While we do have an index on `status`, I am still unsure if the database will using it for a `contains` query and instead perform a full/partial table scan over all the builds while it will definitely be using the index when it's an `exact` query. I tried to look for an explanation but couldn't find much except [this](https://stackoverflow.com/a/2336940/4804277) but newer versions of Postgres claim to have a very fast `LIKE`, so I don't know how much of it is relevant.

Irrespective of the above, matching fields like statuses should be exact anyway in my opinion. In this case, an appropriate value to be searched in the database could be tied to the "All Failed Builds". I need to think a little more on how we can implement it. What do you suggest, @mojca?

ports/templatetags/url_generate.py
arjunsalyan: Sorry, my bad. Improved this.

ports/models.py
```diff
@@ -28,47 +34,398 @@ class Port(models.Model):
...
+            load_dependencies_table(ports)
+
+    @classmethod
+    def update(cls, data, is_json=True):
+        def open_portindex_json(path):
+            with open(path, "r") as file:
+                ports = json.load(file)
+            return ports
+
+        def open_ports_from_list(list_of_ports, path='portindex.json'):
```
umeshksingla, line 157: May I know the use case for this function?

arjunsalyan: The method of using commit hash would give us a list of ports that need to be updated. This function would fetch those ports from PortIndex.json and return a list of JSON objects which can then be passed to `run_updates`.

.travis.yml
arjunsalyan: I don't know but removing `Django` from `requirements.txt` does not seem like a good idea to me. This method ensures that `Django` version installed is the one provided in `$DJANGO_VERSION` and not the one in `requirements.txt`.

I read about it here: https://docs.travis-ci.com/user/languages/python/#testing-against-multiple-versions-of-dependencies-eg-django-or-flask.

parsing_scripts/git_update.py
arjunsalyan: We are checking for the `remote` to be set to `macports/macports-ports`. This test serves dual purpose- presence of `.git` and at the same time ensures that the it is connected to the right remote.

parsing_scripts/git_update.py
arjunsalyan: Okay, I will add more tests and try to do some improvements.

ports/models.py
umeshksingla: Okay, thank you. I do not know the format of `portindex.json` but in order to avoid any delay in me going through the code of `portindex2json.tcl`, I'd request you to check if it's possible to iterate over `list_of_ports` instead and see if a port can be accessed in O(1) from `portindex.json` (which highly depends on how the ports are stored in portindex.json)?

arjunsalyan: The format of `portindex.json`  is a list of JSON objects. After looking into many options, I will the best week can do is list comprehensions.

Here are the formats:
```
list_of_ports = ["portA", "portB", ..... ]
```

portindex.json
```
[
  {
    "name": "portA",
    "version": 1,
    "random_key": "value"
    .
    .
  },
  {
    "name": "portB",
    "version": 2,
    "random_key": "value2"
    .
    .
  },
.
.
.
}]

File: ports/models.py (hunk 45 above)
Comment (arjunsalyan) [reply ports/models.py]: Thank you, probably `full_data` and `add_to_summary` were both redundant.

File: .travis.yml (hunk 12 above)
Comment (umeshksingla) [reply .travis.yml]: You should probably test before considering my comment but I think the version in requirements file will override the version installed before. Are you saying Travis will automatically disregard the `django` inside requirements?

File: parsing_scripts/git_update.py (hunk 13 above)
Comment (umeshksingla) [reply parsing_scripts/git_update.py]: I guess your way is also a little better in cases someone put an empty `.git` directory inside. I am resolving the comment but I came across this module called [GitPython](https://gitpython.readthedocs.io/en/stable/) I wanted to share.

File: ports/models.py (hunk 47 above)
Comment (umeshksingla) [reply ports/models.py:157]: The format of portindex.json output doesn't directly allow us to access a port by its name. While it can still be improved, I'm anyway resolving this conversation because the feature is not likely to be used much I think and we can always improve it later when we incur any delay issues.

File: ports/models.py (hunk 42 above)
Comment (umeshksingla) [reply ports/models.py:138]: These two are not the same. The one here is part of `load` at around line 142 but the link above refers to changes in `update` around line 255.

Comment (arjunsalyan) [reply ports/models.py:138]: Uh Oh! I am sorry. I kept working on `updates`. Fixed.

Comment (umeshksingla) [reply ports/models.py:138]: Thank you.

Comment (umeshksingla) [reply ports/models.py:138]: I would have given it a different name though to keep the definition of `populate` consistent across other models like `BuildHistory`, `Submission`. Not a big deal, resolving the thread.

File: ports/models.py (hunk 44 above)
Comment (arjunsalyan) [reply ports/models.py:151]: `load` is faster than `update` because it does not have to care of any previously available data or relations and thus less number of queries. I had tried to reduce the number of queries for `update` but I believe they should remain more than `load` in any case. Some tasks in `load` are performed using `bulk_create` which significantly reduces the number of queries.

The difference would be only a few minutes on a local database. But with some remote database it might become significant.

File: parsing_scripts/git_update.py (hunk 36 above)
Comment (arjunsalyan) [reply parsing_scripts/git_update.py]: This can be resolved probably.

Comment (umeshksingla) [reply parsing_scripts/git_update.py]: Although the comments do mention `macports-ports` at multiple places in the code, I believe we can live with it since we are not trying to generalize the script anyway.

File: MacPorts/settings.py (hunk 31 above)
Comment (mojca) [reply MacPorts/settings.py:149]: Trying to clarify what I wanted to say here. We could have a master branch, and potentially a "deployment" branch, and these IPs would not be listed in the master branch. I just find them strange to have them here. But let's postpone this. We can delete them later.

File: api_v1/serializers.py [hunk 48]
```diff
@@ -0,0 +1,49 @@
...
+
+    class Meta:
+        model = BuildHistory
+        fields = ('builder_name', 'build_id', 'status', 'time_start', 'time_elapsed', 'watcher_id')
+
+
+class PortListSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Port
+        fields = ('name', )
```
mojca, line 49: Three cases have a trailing comma, three don't. Be consistent. If you put it in a single line, probably leave the comma out.

api_v1/urls.py
```diff
@@ -0,0 +1,12 @@
+from django.urls import path
+from django.conf.urls import url
+
+from . import views
+
+urlpatterns = [
+    path('ports/count/', views.fetch_total_number_of_ports, name='fetch_total_number_of_ports'),
+    path('ports/<int:page_size>/page/<int:page_num>/', views.fetch_paginated_ports, name='fetch_paginated_ports'),
```
mojca, line 8: This API call sounds a bit strange to me (I wanted to comment immediately, and didn't / forgot). I'm not sure what the best way would be. Web sites would use a question mark with optional parameters after it, but I'm not sure if that's also the usual way for such API.

If I look at GitHub's own API:
https://developer.github.com/v3/repos/commits/#list-commits-on-a-repository
they do seem to use question marks in this exact way:
```
https://api.github.com/resource?page=2
```
Maybe something like
```
ports/list?count=100&page=3
```
or
```
# meaning to start with the first port that's alphabetically >= as `pe` (could be perl5)
ports/list?count=30&start_with=pe
```

api_v1/views.py
```diff
@@ -0,0 +1,67 @@
...
+        ports = Port.objects.filter(categories__name__iexact=category).only('name')
+        serializer = PortListSerializer(ports, many=True)
+        return JsonResponse(serializer.data, safe=False)
+    else:
+        return JsonResponse(ERROR405)
+
+
+def fetch_total_number_of_ports(request):
+    if request.method == 'GET':
+        count = Port.objects.all().count()
```
mojca, line 51: Do you actually need to use `.all()` here? Wouldn't `Port.objects.count()` already work? (not sure, just asking, trying to avoid fetching data when not necessary; but maybe the above expression is already efficient enough anyway)

requirements.txt
```diff
@@ -9,27 +10,38 @@ coverage==4.0.3
 cryptography==2.6.1
 DateTime==4.3
 Django==2.1.7
+django-cors-headers==3.0.2
```
mojca, line 13: This comment is not specific to any particular line, but eventually I would like to get rid of so stringent requirements. I would only specify some limitations when there's a known incompatibility and an older version is needed.

With unit tests we should always be testing against the latest version from pypi, so that one can immediately notice if something gets broken and needs fixing for compatibility, else we might end up having troubles as time passes by. Maybe we need to document what version of dependencies was last used & well tested, but I would like to avoid sticking with an ever increasing age of django, for example.

ports/models.py
mojca: I don't particularly like code duplication either, and I'm still confused about the amount of time it takes to populate our database, but as long as we have such a huge gap between update and efficient data store, I don't mind if we keep an efficient method in. Just make sure that it's properly and extensively backed by unit tests. (The number of unit tests is relatively small compared to the amount of code we have so far.)

mojca (commented):
Just some random comments.

I don't have any good suggestion for how to split the CR, other than ... merging it, open new issues for stuff that needs fixing, continue the new code inside a new PR, and we can always come back to this one later.

api_v1/urls.py
arjunsalyan: Thank you, using GET variables definitely seems better.

api_v1/serializers.py
arjunsalyan: This is something with django rest framework. For list/ tuple with just one item- it forces me to use a trailing comma, otherwise throws an error. I tried using using a single item without a tuple, that does not work either.

mojca: Sorry, python docs confirm this.

mojca: Sorry, python docs confirm this.

mojca: Sorry, python docs confirm this.

api_v1/views.py
arjunsalyan: This is probably the recommended way. It does not fetch everything, but runs a `COUNT` query.

mojca: OK.

umeshksingla: Even though the `Port.objects.all().count()` returns only a `QuerySet` which fetches the data lazily but `Port.objects.count()` would work too. I think it really is just about an extra step. Django docs do not talk about the first one though, only the second method. Here: https://docs.djangoproject.com/en/2.1/ref/models/querysets/#count

api_v1/serializers.py
umeshksingla: @arjunsalyan You cannot have a tuple as `x = ('a')` but both `x = 'a',` or `x = ('a',)` are valid, it's not an issue of Django but python forbids it. In case of more than one element, you can have `y = ('a', 'b')` as well as `y = ('a', 'b',)`. In the case of one element, it cannot distinguish this syntax from simply having an extra bracket in arithmetics, e.g., `a = (1)` vs `a = 1`.

A helpful link [here](https://stackoverflow.com/questions/7992559/what-is-the-syntax-rule-for-having-trailing-commas-in-tuple-definitions) and you'll find more information from further googling.

ports/filters.py
umeshksingla: Carried to https://github.com/macports/macports-webapp/issues/23

requirements.txt
umeshksingla: Carried to #25

api_v1/urls.py
umeshksingla: Carried to #26

MacPorts/settings.py
umeshksingla: over to #27

ports/models.py
umeshksingla: Carried to #24

parsing_scripts/git_update.py
umeshksingla: Carried to #28

.travis.yml
umeshksingla: Refer #29

ports/models.py
umeshksingla: Carried to #30

api_v1/views.py
arjunsalyan: Ok, sorry, I just saw the same usage everywhere. But we should go by the docs. Thanks

PR #31, opened 2019-07-09
Restructure Directory

arjunsalyan:


arjunsalyan:
I added migration files to this. They need to be removed.

PR #33, opened 2019-07-11
Run load and update using portindex.json from rsync

arjunsalyan:
This PR is not meant to be merged.


mojca:
What would really help here would be if you created another temporary branch (or put the contents of this one to another one) and cherry-picked just those commits that are relevant for "updating portindex.json from rsync". Some commits are straightforward and could easily be merged, but combined commits don't help that much.

app/ports/models/port.py
```diff
@@ -313,3 +313,8 @@ class Meta:
             models.Index(fields=['github']),
             models.Index(fields=['name', 'domain'])
         ]
+
+
+class Commit(models.Model):
+    hash = models.CharField(max_length=50)
+    updated_at = models.DateTimeField(auto_now=True)
```
mojca, line 320: Moving this part is ok. What's confusing though is that you'll probably store just the latest commit, right? Or will you store more? If you only store the latest commit, the name should be slightly more descriptive. No idea about the best one, but maybe something in the direction of "LastPortInfoUpdate" class with `git_commit_hash` as the first field.

app/ports/management/commands/update-rsync.py
```diff
@@ -0,0 +1,46 @@
+import subprocess
+import json
+
+from django.core.management.base import BaseCommand, CommandError
+
+from parsing_scripts import git_update
+from ports.models import Port, Commit
+
+RSYNC = "rsync://rsync.macports.org/macports//trunk/dports/PortIndex_darwin_16_i386/PortIndex.json"
```
mojca, line 9: First of all, you should be fetching from 18 rather than 16.

I would like to see one single configuration file where this is one of the variables being used.

Over time the origin of this file might change, if nothing else to use a newer OS. If such information is spread in gazzilion of unrelated files, it is super difficult to find it.

app/ports/management/commands/update-rsync.py
```diff
@@ -0,0 +1,46 @@
...
+JSON_FILE = "portindex.json"
```
mojca, line 10: This one is to some extent "a local variable" (though leaving stuff around after being used), so I don't care that much if you leave the name here. What I do care though is to use the above mentioned global configuration file to store path to (semi)temporary files like this one.

app/ports/management/commands/update-rsync.py
```diff
@@ -0,0 +1,46 @@
...
+            ports_to_be_updated = git_update.get_list_of_changed_ports(options['new'], options['old'])
+
+        # Run updates
+        Port.update(ports_to_be_updated, False)
+
+        # Write commit hash to database
+        if options['new'] is False:
+            Commit.objects.create(hash=data['info']['commit'])
+        else:
+            Commit.objects.create(hash=options['new'])
```
mojca, line 46: Don't you only need one single hash with the latest commit taken into account?

app/ports/management/commands/update-rsync.py
```diff
@@ -0,0 +1,46 @@
+import subprocess
```
mojca, line 1: `update-rsync` is a somewhat misleading name. It could be `update-portindex-and-port-data` (not the best name either, but looking for a better proposal).

app/ports/management/commands/update-rsync.py
```diff
@@ -0,0 +1,46 @@
...
+        subprocess.call(['/usr/bin/rsync', RSYNC, JSON_FILE])
+
+        # Open the file
+        with open(JSON_FILE, "r", encoding='utf-8') as file:
+            data = json.load(file)
+
+        if options['new'] is False:
+            ports_to_be_updated = git_update.get_list_of_changed_ports(data['info']['commit'], False)
+        else:
+            ports_to_be_updated = git_update.get_list_of_changed_ports(options['new'], options['old'])
```
mojca, line 37: I do have some hard time following the logic. It kind of looks as if `options['new']` was commit it, but it seems to be a boolean value. It's possible to figure it out eventually, but it could be somewhat clearer.

app/ports/management/commands/update-rsync.py
```diff
@@ -0,0 +1,46 @@
...
+                            default=False,
+                            help="Define a commit till which the update should be processed")
+        parser.add_argument('old',
+                            nargs='?',
+                            default=False,
+                            help="Not recommended. Helps you provide a commit from which update should start")
+
+    def handle(self, *args, **options):
+        # fetch from rsync
+        subprocess.call(['/usr/bin/rsync', RSYNC, JSON_FILE])
```
mojca, line 28: The path to `rsync` should be defined in the above mentioned configuration.
The path to json file should be absolute rather than relative.
I would probably use `-ac` argument to rsync or something similar to sync the files.

app/ports/models/port.py
arjunsalyan: I was earlier thinking of storing all (when running updates every 12 hours), but if we run them more frequently then they will just eat space in the database. I will change the functionality and names. Thanks

app/ports/management/commands/update-rsync.py
arjunsalyan: Thank you. I will change.

app/ports/management/commands/update-rsync.py
arjunsalyan: If a commit hash is not supplied to the command then `options['new']` is `False` by default.

app/ports/management/commands/update-rsync.py
arjunsalyan: OKay, thanks

arjunsalyan:
> What would really help here would be if you created another temporary branch (or put the contents of this one to another one) and cherry-picked just those commits that are relevant for "updating portindex.json from rsync". Some commits are straightforward and could easily be merged, but combined commits don't help that much.

Okay, I will move the contents which do not have changes in docker config to some other branch and open a PR with it.

app/ports/models/port.py
arjunsalyan: This has been fixed: https://github.com/macports/macports-webapp/commit/ea2eccbaf84e98737dd696e5fc9e0c8c9eb7f7c8#diff-4b58a59fb2052fa5c6ba6dd6a385cf44R318

arjunsalyan:
I have opened  new PR with only the selected commits. I have tried to reference your reviews from this one, sorry, I did not know the right way to do this.

arjunsalyan:
This can be closed now. New PR: https://github.com/macports/macports-webapp/pull/34

PR #34, opened 2019-07-12
Optimise Initial Data Loading and Enable Updates

arjunsalyan:
- [x] Optimise the app to use new portindex.json format
- [x] Add command to load initial data using rsync
- [x] Add command to update using rsync


app/parsing_scripts/rsync_operations.py
```diff
@@ -0,0 +1,20 @@
+import subprocess
+import os
+import json
+
+RSYNC = '/usr/bin/rsync'
```
arjunsalyan: https://github.com/macports/macports-webapp/pull/33#discussion_r302761873: Used all the variables in a single file.

https://github.com/macports/macports-webapp/pull/33#discussion_r302764296: Used RSYNC variable

app/ports/management/commands/update-portinfo.py
```diff
@@ -0,0 +1,50 @@
+from django.core.management.base import BaseCommand, CommandError
```
arjunsalyan, line 1: https://github.com/macports/macports-webapp/pull/33#discussion_r302763269 I have used the name `update-portinfo`.

app/ports/management/commands/update-portinfo.py
```diff
@@ -0,0 +1,50 @@
...
+                            nargs='?',
+                            default=False,
+                            help="Define a commit till which the update should be processed")
+        parser.add_argument('old',
+                            nargs='?',
+                            default=False,
+                            help="Not recommended. Helps you provide a commit from which update should start")
+
+    def handle(self, *args, **options):
+        # Fetch from rsync
```
arjunsalyan: https://github.com/macports/macports-webapp/pull/33#discussion_r302764260 I have added comments for steps to make the logic a bit more clear.

app/parsing_scripts/git_update.py
```diff
@@ -49,8 +49,8 @@ def get_list_of_changed_ports(new_hash=False, old_hash=False, root=BASE_DIR):
             if old_hash is False:
                 try:
                     # Try to fetch the most recent hash from database
-                    old_hash_object = Commit.objects.all().order_by('-updated_at').first()
-                    old_hash = old_hash_object.hash
+                    old_hash_object = LastPortIndexUpdate.objects.all().first()
+                    old_hash = old_hash_object.git_commit_hash
                 except AttributeError:
                     # If database is empty, use the current HEAD
```
mojca: This logic kind of doesn't make sense. The old hash would be the latest commit? If anything, I assume you should start from the first commit instead :)

app/ports/management/commands/load-rsync.py
```diff
@@ -0,0 +1,29 @@
...
+
+    help = "Populates the database with Initial data after fetching portindex.json from rsync"
+
+    def handle(self, *args, **options):
+        if not Port.objects.count() > 0:
+            # Fetch from rsync
+            rsync_operations.sync()
+
+            # Open the file
+            data = rsync_operations.open_file()
```
mojca: What does "open file" have to do with "rsync operations"? The name rsync is super misleading: we could use rsync for just about anything.

You could have a class called `PortIndex` and then call `sync()` or `open()` on that one. And the `sync()` function could potentially call functionality from `RsyncHandler` class if necessary, but in that case `PortIndex` would specify what to sync from where, and the handler would only make sure of transferring the files.

app/ports/management/commands/load-rsync.py
```diff
@@ -0,0 +1,29 @@
...
+
+            # Start populating
+            Port.load(data['ports'])
+
+            # Write the commit hash into database
+            last_commit = LastPortIndexUpdate.objects.all().first()
+            last_commit.git_commit_hash = data['info']['commit']
+            last_commit.save()
```
mojca: While you can leave this as it is for a little while, this field should not be updated unless database population is known to be successful.

What would make slightly more sense to me would be to pass the full json to `Port.load` and then the commit would be updated at the end of that function rather than here. @umeshksingla?

app/ports/management/commands/load-rsync.py
```diff
@@ -0,0 +1,29 @@
...
+    def handle(self, *args, **options):
+        if not Port.objects.count() > 0:
+            # Fetch from rsync
+            rsync_operations.sync()
+
+            # Open the file
+            data = rsync_operations.open_file()
+
+            # Start populating
+            Port.load(data['ports'])
```
mojca: Does it make sense to use some kind of `initial` prefix for the function name? (inital_load, first_load, first_population, initial_population, ... I don't like any of my own proposals here, but I'm sure we can come up with something that's self-explanatory.)

app/parsing_scripts/rsync_operations.py
mojca: You now use a variable, but these variables should be defined together in some "common settings" file.

app/ports/management/commands/load-rsync.py
```diff
@@ -0,0 +1,29 @@
...
+
+            # Write the commit hash into database
+            last_commit = LastPortIndexUpdate.objects.all().first()
+            last_commit.git_commit_hash = data['info']['commit']
+            last_commit.save()
+
+            print("Initial Data Loaded Successfully.")
```
mojca: It doesn't relate to this line in particular, but I didn't know where to write.

To me the filename `load-rsync` is very misleading. It should be clear that it's about "initial port database population".

app/ports/management/commands/load-rsync.py
```diff
@@ -0,0 +1,29 @@
...
+        else:
+            print("The database is not empty. Command cannot run.")
```
mojca: Shouldn't there also be a non-zero return or something?

mojca (commented):
The function get_list_of_changed_ports (which was there before) is still somewhat difficult to follow, in particular with its seemingly recursive calls.

mojca:
I left a bunch of random reviews. This code could use some nice refactoring (but it's also about the code that has been written before). I'm not strongly opposing writing the issues down and addressing some of them later.

app/parsing_scripts/git_update.py
arjunsalyan: `git pull` runs after this step. So the commit before `git pull` is the `old_hash` and the commit after `git pull` is the new hash. I am sorry, I could not get what do you mean by first commit?

This happens only in that case when the database is empty all well as a commit is not supplied.

app/ports/management/commands/load-rsync.py
arjunsalyan: Yes, thank you. The naming here could be improved.

app/ports/management/commands/load-rsync.py
arjunsalyan: Okay, thank you. `load-initial-data` could be the name of the file, as well as the function name suggested by you above.

app/parsing_scripts/git_update.py
mojca: If the database is empty, it could be that no ports exist in the database, to start with. So you could start with commit 1 in the git history, or basically the equivalent of updating all ports :)

I clearly missed the fact that this happens before you fetch contents from git. It's still not optimal, but well ...

arjunsalyan: Okay, thank you. Using the first commit makes sense.

app/ports/management/commands/load-rsync.py
arjunsalyan: Thank you, but if `Port.load()` throws an error, the command will not proceed either.

app/ports/management/commands/load-rsync.py
arjunsalyan: I have used `autoload_initial_data` because there is already `load` which needs to supplied a file manually.

app/MacPorts/config.py
```diff
@@ -0,0 +1,10 @@
+import os
+
+# configuration for rsync
+RSYNC = '/usr/bin/rsync'
+RSYNC_SOURCE = "rsync://rsync.macports.org/macports//trunk/dports/PortIndex_darwin_18_i386/PortIndex.json"
+JSON_FILE = os.path.join(os.path.abspath(os.path.dirname(__name__)), 'portindex.json')
+
+# configuration for repository to find updated ports
+REPO = "macports-ports"
+REPO_URL = "https://github.com/macports/macports-ports.git"
```
umeshksingla: maybe `REPO_URL = "https://github.com/macports/" + REPO + ".git"`?

app/parsing_scripts/git_update.py
```diff
@@ -15,45 +16,43 @@
...
-                    # Try to fetch the most recent hash from database
-                    old_hash_object = Commit.objects.all().order_by('-updated_at').first()
-                    old_hash = old_hash_object.hash
-                except AttributeError:
-                    # If database is empty, use the current HEAD
-                    old_hash = subprocess.run(['git', 'rev-parse', 'HEAD'], stdout=subprocess.PIPE).stdout.decode(
+
+                # If the database has old commit, use it for old_hash
+                if LastPortIndexUpdate.objects.count() > 0:
+                    old_hash_object = LastPortIndexUpdate.objects.all().first()
```
umeshksingla: `.first()` returns `None` if there is no object for a model in db. I think we can remove the need for `.count()` here.
[Link](https://docs.djangoproject.com/en/2.2/ref/models/querysets/#first)

app/parsing_scripts/git_update.py
```diff
@@ -15,45 +16,43 @@
...
+                    old_hash = old_hash_object.git_commit_hash
+                else:
+                    # If database is empty, use the first commit
+                    old_hash = subprocess.run(['git', 'rev-list', 'HEAD', '|', 'tail', '-n', '1'], stdout=subprocess.PIPE).stdout.decode(
```
umeshksingla: At some point, I would like to get rid of all these hacks and use something like [GitPython](https://gitpython.readthedocs.io/en/stable/) as I already mentioned before. But I checked and there seems to be a learning curve for using that module and it is of quite a low priority, so we can probably postpone.
@mojca Should I create an Issue to deal with this later or we can live with this?

app/parsing_scripts/git_update.py
```diff
@@ -73,21 +72,19 @@ def get_list_of_changed_ports(new_hash=False, old_hash=False, root=BASE_DIR):
             for line in s:
                 portname = line.split('/')[1]
                 if portname not in updated_ports:
-                    updated_ports.append(portname)
+                    updated_ports.append(portname.lower())
```
umeshksingla: I may have not followed the conversation but why are we converting it to lower case?

app/parsing_scripts/git_update.py
umeshksingla: Also, can you check if `.all()` is needed before `.first()`?

umeshksingla (commented):
The `update` part seems like it can use some serious refactoring. Let's resolve some of the comments first and then get to it. Overall, the PR seems good to me.

app/parsing_scripts/git_update.py
arjunsalyan: When I wrote this script, I did check for any modules easing out the process. I also checked `GitPython`, and I remember it did not support all the functionality we need. We can definitely use GitPython, but probably it won't be able to replace all commands.

app/parsing_scripts/git_update.py
arjunsalyan: I did this just to avoid any missed ports because of case. Port names are case insensitive and at some places like `dependencies`, some maintainers have not been consistent with case. So, I wanted to avoid that situation here. Could we improve this in some other way?

app/parsing_scripts/git_update.py
arjunsalyan: > `.first()` returns `None` if there is no object for a model in db. I think we can remove the need for `.count()` here.
> [Link](https://docs.djangoproject.com/en/2.2/ref/models/querysets/#first)

And then use `if-else` to check if it is `None`?

app/MacPorts/config.py
arjunsalyan: Thank you. This would be better.

app/parsing_scripts/git_update.py
umeshksingla: Yes. I mean I agree it's a really really small improvement but ...

app/ports/management/commands/update-portinfo.py
```diff
@@ -37,13 +37,24 @@ def handle(self, *args, **options):
...
+        portnames_found_in_portindex = []
         for port in data['ports']:
             if port['name'].lower() in ports_to_be_updated:
                 ports_to_be_updated_json.append(port)
+                portnames_found_in_portindex.append(port['name'].lower())
+
+        # Find deleted ports
+        deleted_ports = []
+        for portname in ports_to_be_updated:
+            if portname.lower() not in portnames_found_in_portindex:
```
mojca: This could easily be O(n^2) if you happen to update all ports (which did happen in the past, for example when the svn expansion was removed). Super inefficient, don't do it.

If you do this, at least use a dictionary rather than an array to check whether a port exists in a set.

That said, did you happen to notice that if your input data structure (at the moment `data['ports']`) was itself a dictionary, it would be sufficient to just check
```python
if port['name'] in data['ports']
```
as opposed to having to first transform and store the data in a different way?

I need to collect some thoughts to decide if I'm happy with the way this is done. I believe it's time to create unit tests for this part of functionality, as this might get tricky. A bunch of questions:
* The answer might be a provisional no, but could it be that somehow a list of ports that need as update accidentally contains a new port that was added one commit after the portindex was generated? Will you ever set the status of that port back to active / where?
* What happens when just a subport is removed?
* What happens when the full Portfile is deleted?
* What happens when a (sub)port is moved from one file to another?
* Does it work correctly when a port with the same name is added after a while (will it be set to active again)?

Could we cover these scenarios in unit tests?
I also need to check again the distinction between deleted and obsolete ports.

app/MacPorts/config.py
```diff
@@ -0,0 +1,10 @@
+import os
+
+# configuration for rsync
+RSYNC = '/usr/bin/rsync'
+RSYNC_SOURCE = "rsync://rsync.macports.org/macports//trunk/dports/PortIndex_darwin_18_i386/PortIndex.json"
```
mojca: Call this rather `PORTINDEX_SOURCE` or `PORTINDEX_RSYNC_SOURCE` or `PORTINDEX_JSON_RSYNC_SOURCE` or something like that.
Also remove double slash.

app/MacPorts/config.py
```diff
@@ -0,0 +1,10 @@
+import os
+
+# configuration for rsync
+RSYNC = '/usr/bin/rsync'
```
mojca: Maybe also add `GIT`?

app/MacPorts/config.py
```diff
@@ -0,0 +1,10 @@
...
+RSYNC_SOURCE = "rsync://rsync.macports.org/macports//trunk/dports/PortIndex_darwin_18_i386/PortIndex.json"
+JSON_FILE = os.path.join(os.path.abspath(os.path.dirname(__name__)), 'portindex.json')
```
mojca: * Call stuff after contents, not just by type, unless it's a super local variable. Something like `PORTINDEX_JSON` maybe.
* If you do it this way, use a reasonable name for `os.path.abspath(os.path.dirname(__name__))` and do the `os.path.join` between that variable and `portindex.json`. I would probably create a separate folder outside of the sources themselves and store git clones and stuff like that all at that same place. It can be `data` next to `app`, it doesn't need to be stored in another universe, but a path like the one above definitely sounds strange.

app/MacPorts/config.py
```diff
@@ -0,0 +1,10 @@
...
+
+# configuration for repository to find updated ports
+REPO = "macports-ports"
+REPO_URL = "https://github.com/macports/" + REPO + ".git"
```
mojca: I don't super agree with @umeshksingla about splitting the repository URL this way. One repository usually has one path, it can be `ssh` path to some other exotic place.

This splitting would make sense if you need to use 10 repositories from the same organisation throughout your app. Here I see little to no sense in doing this splitting. But more importantly, the name `REPO` is not too descriptive when put to a separate configuration file (it could be repo of webapp itself, for example). How about the following?
```
MACPORTS_PORTS_URL = "https://github.com/macports/macports-ports.git"
```

I later noticed that you use `REPO` in `os.path.isdir(config.REPO)` or `REPO_DIR = os.path.join(root, config.REPO)` (which is also a tiny bit inconsistent).

Just define some kind of `DATA_DIR` (feel free to use a better name than that) and something like
```
MACPORTS_PORTS_DIR = os.path.join(DATA_DIR, 'macports-ports')
```
and also use this in
```
git clone $MACPORTS_PORTS_URL $MACPORTS_PORTS_DIR
```
(with proper quoting, the above is just pseudocode).

app/parsing_scripts/git_update.py
mojca: I don't really like the idea about introducing another heavy dependency with an additional abstract layer of stuff that hinders our understanding of the code and makes it more difficult to write new queries (like Django makes it super difficult to work with somewhat more complex join queries).

My only request would be to define path to `git` next to path to `rsync`.

I would leave the rest intact.

app/ports/tests/test_git_update.py
```diff
@@ -14,13 +15,15 @@ def test_between_new_old(self):
...
-        self.assertEquals(ports, ['ideviceinstaller', 'OpenBLAS', 'nomad'])
+        self.assertEquals(ports, ['ideviceinstaller', 'openblas', 'nomad'])
 
     def test_between_new_and_database(self):
-        git_update.get_list_of_changed_ports("7646d61853154b9dc523e3d0382960aea562e7ab",
-                                             "ee0fc9c4ca59685f33628fc41c6946920869ab71",
-                                             TEST_REPO_PARENT
-                                            )
+        if LastPortIndexUpdate.objects.count() > 0:
+            last_commit = LastPortIndexUpdate.objects.all().first()
```
mojca: Would something like this work?
```
last_commit = LastPortIndexUpdate.objects.all().first()
if last_commit != nil
```
Or maybe even better: could `update_or_create()` work at this place?

We probably also need something to enforce a single entry at the database level. (Random link with an idea: https://stackoverflow.com/a/25393923, even if not so extremely important at this very point.)

app/ports/tests/sample_data/portindex2json.tcl
```diff
@@ -2,8 +2,11 @@
 # a list of JSON objects.
 # Written by Joshua Root <jmr@macports.org>, 2017
 # Requires: tclsh with the tcllib 'json' package.
-# Usage: tclsh portindex2json.tcl path/to/PortIndex
-
```
mojca, line 6: Oh, wait! Couldn't/shouldn't we clone the script from the upstream repository instead? It just feels wrong to keep a literally copy here.

app/parsing_scripts/git_update.py
mojca: First: use dictionary rather than list, else you'll end up in quadratic time (with 30.000 ports and all of them updated, you can easily generate a billion string comparisons).

Second (even more important): this command gives you PATHS!!! rather than ports that need updating. If one changes `perl5`, any of its subports may be affected (`perl5.30`, `perl5.28`, ...), not just the main one.

Important:
* You need an unit test to check what happens when directory is renamed from lowercase to uppercase.
* You need an unit test to ensure that subports all get updated / removed / obsoleted etc. when something changes.

app/MacPorts/config.py
arjunsalyan: Yes, thank you.

app/MacPorts/config.py
arjunsalyan: Thank you, idea of a `data` folder would make things neat.

app/ports/management/commands/update-portinfo.py
arjunsalyan: I have now made use of a dictionary.


> * The answer might be a provisional no, but could it be that somehow a list of ports that need as update accidentally contains a new port that was added one commit after the portindex was generated? Will you ever set the status of that port back to active / where?

If due to some mess with git commits, a port gets marked as inactive, it will remain inactive until a change is made to the portfile and the app gets an opportunity to update it.

> * What happens when just a subport is removed?

The current functionality will detect when a support is deleted. I have also added a simple unitest for same, which can be made more complex.

> * What happens when the full Portfile is deleted?

Main port and all subports will be marked as deleted.

> * What happens when a (sub)port is moved from one file to another?

The previous port will be marked as deleted and new one will be created.

> * Does it work correctly when a port with the same name is added after a while (will it be set to active again)?

Yes, I have added a unitest also.

> Could we cover these scenarios in unit tests?

I have already covered only some, will cover all very soon.

app/MacPorts/config.py
arjunsalyan: I have made suggested changes.

app/MacPorts/config.py
arjunsalyan: I have probably addressed this, thank you.

app/ports/tests/test_git_update.py
arjunsalyan: I have created a method `update_or_create_first_object` inside the Model class.

app/ports/tests/sample_data/portindex2json.tcl
arjunsalyan: Okay, thank you.

app/parsing_scripts/git_update.py
```diff
@@ -18,54 +18,50 @@
 
 
 def clone_repo():
-    if os.path.isdir(config.REPO):
+    os.chdir(config.DATA_DIR)
+    if os.path.isdir(config.MACPORTS_PORTS_DIR):
         return
     else:
-        os.system("git clone %s" % config.REPO_URL)
+        subprocess.run([config.GIT, 'clone', config.MACPORTS_PORTS_URL])
```
mojca: For total consistency reasons (not really important though, result would currently stay the same) you would add another parameter `config.MACPORTS_PORTS` as destination for cloning.

app/parsing_scripts/git_update.py
```diff
@@ -18,54 +18,50 @@
...
 
 
-def get_list_of_changed_ports(new_hash=False, old_hash=False, root=BASE_DIR):
-    os.chdir(root)
+def get_list_of_changed_ports(new_hash=False, old_hash=False):
+    if os.path.isdir(config.MACPORTS_PORTS_DIR):
```
mojca: This is a bit of nitpicking, but when code consists of a super log portion with the main loop + a tiny bit of code to handle the `else` part, it's more difficult to follow the flow (and even find the `else` part). What you can do is start with the short else part and immediately return. Then for the main part you don't even need to use extra alignment. I can tell more in a chat if needed.

README.md
```diff
@@ -30,14 +30,14 @@ mentored by the [MacPorts](https://www.macports.org) organisation.
...
 # Running the App
-The repository contains Docker configuration which can be used out of the box, however, the app can be run without using
-the docker container. Both way are discussed below:
+The repository contains Docker configuration which can be used to build docker image, however, the app can be run without using
+the docker container. Both ways are discussed below:
 
 ## 1. Run inside Docker Container
 It is the recommended way, the Docker Image of the app is pre-configured to run the `migrations` and `collectstatic`.
-However, the `load` command to populate the database and the crons need to be run manually.
+However, the commands to populate the database with initial data and start the cron jobs need to be run manually.
```
mojca:
```suggestion
However, the commands to populate the database with the initial data and to start the cron jobs need to be run manually.
```

README.md
```diff
@@ -30,14 +30,14 @@ mentored by the [MacPorts](https://www.macports.org) organisation.
 ___
 
 # Running the App
-The repository contains Docker configuration which can be used out of the box, however, the app can be run without using
-the docker container. Both way are discussed below:
+The repository contains Docker configuration which can be used to build docker image, however, the app can be run without using
```
mojca:
```suggestion
The repository contains Docker configuration which can be used to build the docker image, however, the app can be run without using
```

README.md
```diff
@@ -60,26 +60,49 @@ DB_NAME=
...
 ```
-docker run -d -p 80:80 --env-file=env macports-webapp
+docker run -d -p 80:80 --env-file=env --name=macports-webapp macports-webapp
 ```
+and visit `127.0.0.1:80` to see if everything looks fine. Sometimes, it may take 10-15 seconds for the site to display
+the layout properly because `migrations` and `collectstatic` commands are running in the background. There will not be any
+contents in the site because the database is empty, but it should load up nicely with empty sections.
 
-To start the cron jobs:
+2. Populate the database with initial data using the command:
```
mojca:
```suggestion
2. Populate the database with the initial data using the command:
```

app/parsing_scripts/git_update.py
```diff
@@ -6,88 +6,83 @@
...
-            print("Cleaning current tree and cloning new repo.")
-            shutil.rmtree(REPO)
-            clone_repo()
-            return get_list_of_changed_ports(new_hash, old_hash, root)
-    else:
-        print('{} directory not found. Cloning into'.format(REPO))
+                old_hash = old_hash_object.git_commit_hash
+
+        # Pull from macports-ports
+        subprocess.call([GIT, 'pull'])
```
mojca, line 60: This would be ok for now, but in case something goes wrong, it is not super reliable.

mojca (commented):
Just some trivial comments before I fall asleep, I'll continue tomorrow.

README.md
arjunsalyan: Thank you. Would you mind if I create a single commit with all these improvements?

config/crons
```diff
@@ -2,3 +2,5 @@ SHELL=/bin/bash
 BASH_ENV=/code/cron.env
 
 0 * * * * root python3 /code/app/manage.py fetch-build-history >> /var/log/buildhistorycron.log
+
+0 */3 * * * root python3 /code/app/manage.py update-portinfo >> /var/log/updatecron.log
```
mojca, line 6: We could later switch to more frequent updates than every three hours, but it looks ok for now. It's already waaaay more helpful to have this working :)

app/ports/tests/test_git_update.py
```diff
@@ -1,37 +1,30 @@
...
-                                                     TEST_REPO_PARENT
-                                                     )
-        self.assertEquals(ports, ['nomad', 'terraform'])
+        ports = git_update.get_list_of_changed_ports("f3527caa506b1b944f43f47085dd35a8b8e2b050")
+        self.assertEquals(ports, {'net/nomad': '', 'sysutils/terraform': ''})
 
     def test_broken_repo(self):
-        broken_repo_parent = os.path.join(TEST_REPO_PARENT, 'broken_directory')
+        os.chdir(MACPORTS_PORTS_DIR)
+        shutil.rmtree('.git')
```
mojca, line 26: Are you removing the same `.git` folder that's used for real port updates?

If yes, what happens if you run the tests at the same time when another cron job is running?

app/ports/models/port.py
```diff
@@ -27,15 +30,28 @@ class Port(models.Model):
     name = models.CharField(max_length=100, db_index=True)
     license = models.CharField(max_length=100, default='')
     replaced_by = models.CharField(max_length=100, null=True)
+    active = models.BooleanField(default=True)
 
     objects = PortManager()
 
+    class RsyncHandler:
```
mojca: You could call this `PortIndexUpdateHandler`. The name `RsyncHandler` is too generic.

app/ports/models/buildhistory.py
```diff
@@ -97,9 +99,29 @@ def load_database(data):
             if build_number_loaded:
                 build_in_database = build_number_loaded[0].build_id + 1
             else:
-                build_in_database = last_build_number - 9999
+                build_in_database = last_build_number - 20
```
mojca: Can we turn this number into a constant (and put it to `config.py`)?

app/ports/management/commands/update-portinfo.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def handle(self, *args, **options):
+        # Fetch from rsync
+        Port.RsyncHandler().sync()
+
+        # Open the file
+        data = Port.RsyncHandler().open_file()
+
+        # If no argument is provided, use the commit-hash from JSON file:
+        if options['new'] is False:
```
mojca: What if we call the option `new_commit` instead of `new`, and use `None` instead of `False`?

app/ports/management/commands/update-portinfo.py
```diff
@@ -0,0 +1,57 @@
...
+            new_hash = data['info']['commit']
```
mojca: And maybe also call this variable `new_commit`.

app/ports/management/commands/update-portinfo.py
```diff
@@ -0,0 +1,57 @@
...
+        # If argument is provided
+        else:
+            new_hash = options['new']
+
+        # Using the range of commits, find the list of ports which have been updated
```
mojca: ... updated between the given two commits.

app/ports/management/commands/update-portinfo.py
```diff
@@ -0,0 +1,57 @@
...
+        # options['old'] will be false if no argument is provided
```
mojca: If no argument for `old_commit` was provided, `options['old_commit']` will be `None` and the code will use the last commit from the previous update.

(Just suggestion.)

app/ports/management/commands/update-portinfo.py
```diff
@@ -0,0 +1,57 @@
...
+                found_ports_tree[portdir][portname] = True
+
+        # Mark deleted ports
+        Port.mark_deleted(found_ports_tree)
+
+        # Run updates
+        Port.update(ports_to_be_updated_json)
+
+        # Write the commit hash into database
+        LastPortIndexUpdate.update_or_create_first_object(data['info']['commit'])
```
mojca, line 54: Do you still update the last commit if the update action fails for some reason?

app/ports/management/commands/autoload-initial-data.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def handle(self, *args, **options):
+        if Port.objects.count() or Builder.objects.count() > 0:
+            raise CommandError("The database is not empty. Command cannot run.")
+
+        # Fetch from rsync
+        Port.RsyncHandler().sync()
+
+        # Open the file
+        data = Port.RsyncHandler().open_file()
```
mojca, line 18: In a way it could make sense to run `sync_and_open_file()` (just a single command doing both), but it's also ok as it is.

app/ports/tests/test_git_update.py
arjunsalyan: Oh, I totally forgot that we might be running tests at places other than Travis also. This needs to be improved.

But, then what could be the right way to specify the command to use a different directory? It uses the path from `config.py`.

app/ports/models/port.py
arjunsalyan: Okay, thanks.

app/ports/models/buildhistory.py
arjunsalyan: That would be better, thanks.

app/ports/management/commands/update-portinfo.py
arjunsalyan: No. If the update action fails, I believe the command will not proceed.

mojca:
I renamed the filename of migration. I hope that's ok.

PR #42, opened 2019-07-18
Add project intro to README.md

### LINKED ISSUE #41: README.md does not describe what the code does
The top level README.md explains how to run the webapp but not what it is. A few paragraphs of explanation at the beginning would be in order.

arjunsalyan:
Closes: https://github.com/macports/macports-webapp/issues/41

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,34 @@
+# Introduction
+
+### Goal
+
+This is a web application that:
+ 1. Displays general information about all the ports.
+ 2. Fetches and displays build history of the ports.
+ 3. Collects and displays installation statistics (system configurations and installed ports) from users who opt-in by
+    installing the port `mpstats-gsoc`.
+    
+It aims at being an all-one-place for users to find relevant ports and for maintainers to monitor their ports and
+understand their user base.
+ 
+More info at [docs/1-INFORMATION.md](docs/1-INTRODUCTION.md)
+
+### Architecture
+
+The web application is based on the [Django Framework](http://djangoproject.com) utilising a PostgreSQL database.
+
+The app is supposed to be deployed in a docker container with nginx and uWSGI serving the content.
+
+**Demo**: [AWS EC2](http://ec2-52-34-234-111.us-west-2.compute.amazonaws.com) (running a docker container)
+ 
+This web application has been implemented during the [Google Summer of Code 2019](https://summerofcode.withgoogle.com) 
+mentored by the [MacPorts](https://www.macports.org) organisation.
+
+**Student**: [Arjun Salyan](https://github.com/arjunsalyan) <br>
+**Mentors**: [Mojca Miklavec](https://github.com/mojca), [Umesh Singla](https://github.com/umeshksingla) <br>
+
+___
+
 # Running the App
 The repository contains Docker configuration which can be used out of the box, however, the app can be run without using
 the docker container. Both way are discussed below:


README.md
```diff
@@ -1,3 +1,28 @@
+# 1. Introduction
+
+This web application is a [Google Summer of Code 2019](https://summerofcode.withgoogle.com) project under the [MacPorts](https://www.macports.org) organisation.
```
mojca: I would probably say something more in the sense of "This web application has been implemented during the [Google Summer of Code 2019](https://summerofcode.withgoogle.com) as a project mentored by the [MacPorts](https://www.macports.org) organisation."

But slightly more important it might make sense to start with the "what this is" part rather than how it came to be.

README.md
```diff
@@ -1,3 +1,28 @@
+# 1. Introduction
```
mojca: I would probably leave the number 1 out.

README.md
```diff
@@ -1,3 +1,28 @@
...
+
+The web application is based on the [Django Framework](http://djangoproject.com) utilising a PostgreSQL database.
+
+The app is supposed to be deployed in a docker container with nginx and uWSGI serving the content.
+
+**Demo**: [AWS EC2](http://ec2-52-34-234-111.us-west-2.compute.amazonaws.com) (running a docker container)
+
+### Goal of the Project
+
+Build an all-in-one application that:
```
mojca: I would make this the first paragraph. I would not start as "Build an ...", but rather something like "This is a web application which:". The points below should probably all start with a capital letter (I have a feeling these are some strange rules of the English grammar, but I'm not absolutely sure).

README.md
```diff
@@ -1,3 +1,28 @@
...
+ 1. displays general information about all the ports
+ 2. fetches and displays build history about the ports
+ 3. collects data from users and displays installation statistics of ports along with the system configuration of users.
+ 
+NOTE: More info at `docs/1-INFORMATION.md`
```
mojca: Make this a working link. I believe you can make a relative link without having to write full URL to the repository.

README.md
```diff
@@ -1,3 +1,28 @@
+# 1. Introduction
+
+This web application is a [Google Summer of Code 2019](https://summerofcode.withgoogle.com) project under the [MacPorts](https://www.macports.org) organisation.
+
+**Student**: [Arjun Salyan](https://github.com/arjunsalyan) <br>
+**Mentors**: [Mojca Miklavec](https://github.com/mojca), [Umesh Singla](https://github.com/umeshksingla) <br>
+
+### Architecture of the Project
```
mojca: This could be the second paragraph. The last one could be about who wrote the app.

README.md
```diff
@@ -1,3 +1,28 @@
...
+The app is supposed to be deployed in a docker container with nginx and uWSGI serving the content.
+
+**Demo**: [AWS EC2](http://ec2-52-34-234-111.us-west-2.compute.amazonaws.com) (running a docker container)
+
+### Goal of the Project
+
+Build an all-in-one application that:
+ 1. displays general information about all the ports
+ 2. fetches and displays build history about the ports
+ 3. collects data from users and displays installation statistics of ports along with the system configuration of users.
```
mojca: Maybe something like "Collects and displays installation statistics (system configurations and installed ports) from users who opt-it." I would make it clear that user need to explicitly consent rather than having MacPorts doing something on its own.

mojca (commented):
You should add a line
```
Closes: https://github.com/macports/macports-webapp/issues/41
```
or
```
Closes: #41 
```
which will automatically close and properly reference the issue when this gets merged. We usually use the longer form because of the need to reference trac tickets.

README.md
arjunsalyan: Shall I drop the headings?

mojca: It's up to you. Both ways is probably fine.

arjunsalyan:
> You should add a line
> 
> ```
> Closes: https://github.com/macports/macports-webapp/issues/41
> ```
> 
> or
> 
> ```
> Closes: #41 
> ```
> 
> which will automatically close and properly reference the issue when this gets merged. We usually use the longer form because of the need to reference trac tickets.

Thank you.

arjunsalyan:
I have probably address most of the feedback. Thank you so much.

README.md
the-godess: 306daae528dc137c9053554c45e90a631ef859490a3ede651d488135602500a3

Issue #39, opened 2019-07-13
Create a single paginated table with port installation statistics


mojca:

mojca:
Very cool!

What I miss most is some "accumulated sorting" (not sure how to call this). That is: when you first click that you want to sort by total number of installations, then you click requested number of installations, it could make sense to sort by requested number first. Then among the ports with the same number of requested installations, sort by the number of total installations. And finally by port name. In order to do this you might need to add some multi-indices to tables, and you would need to keep track of how sorting was requested. We only have three columns, so you would only need to remember the order of three values. Default sorting order could be (requested desc, total desc, name asc).

One personal request would be to add the display parameters to URL, so that clicking back in the browser would get you back to the previous view. Low priority. Something like `statistics/ports/?days=30,page=2,orderby=requested_desc-total_desc-name_asc`. The last one could be coded in a more compact way like `orderby=RTn` or anything else.

In the future it would be nice to see the following added to the table:
* a column indicating port health summary (it could be just a single symbol)
* a column indicating whether the port is outdated
* a separate filter, similar to "filter by port name" where one could filter by maintainer; then I could quickly check popularity, health status and outdatedness of all my ports

arjunsalyan:
One problem I am facing with multi-field sorting is  priority- which fields to give priority while sorting. If I use the field selected first as the higher priority, then suppose the user wants to change the priority? What is the best way to allow this? A reset button?

mojca:
I would start with default being
```
(requested DESC, total DESC, name ASC)
```
Now assume the user clicks on `requested ASC`. Then you change to
```
(requested ASC, total DESC, name ASC)
```
Now the user clicks on `total DESC`. You change the order into
```
(total DESC, requested ASC, name ASC)
```
Then the user clicks `name DESC`. So you change the order into
```
(name DESC, total DESC, requested ASC)
```
Basically you always keep the set of all three variables (and ideally also reflect those changes in page URL) and whenever the user clicks one of them, push that one to the front with the requested sorted direction. You don't really need to reset at any point (unless when loading a page again, in which case you would set the sorting order to default), you just keep repeating this cycle ad-infinity. There are only 6 permutations of the order (times 8 different sorting direction combinations).

I don't know whether you can create indices for these kind of searches (probably not unless you use materialized views).

This is just a random thought, I'm not saying it's the ideal solution. Did you now approximately understand what I wanted to suggest? Do you have any better ideas?

[...]

PR #43, opened 2019-07-18
Workspace

arjunsalyan:
Apart from some new functionalities most of these commits are template based improvements.


app/ports/views.py
```diff
@@ -151,18 +151,28 @@ def portdetail_build_information(request):
 
 
 def portdetail_stats(request):
+    days = request.GET.get('days', 30)
+    days_ago = request.GET.get('days_ago', 0)
+    try:
+        days = int(days)
+        days_ago = int(days_ago)
```
mojca: You may also want to check that these are not negative?

app/ports/templates/ports/port-detail/installation_stats.html
```diff
@@ -1,5 +1,24 @@
...
 
+<br>
+<form class="bg-light p-2 form form-inline justify-content-center">
+    <lable for="days">Displaying statistics for users who made submissions in previous:</lable>
+    <select onchange="installationStatsAjax()" id="days" name="days" class="format-control mx-2">
+        <option value="7" {% if days == 7 %}selected{% endif %}>7 Days</option>
+        <option value="30" {% if days == 30 %}selected{% endif %}>30 Days</option>
+        <option value="90" {% if days == 90 %}selected{% endif %}>90 Days</option>
+        <option value="180" {% if days == 180 %}selected{% endif %}>180 Days</option>
+        <option value="365" {% if days == 365 %}selected{% endif %}>365 Days</option>
```
mojca: This kind of repetition seems a bit error prone when you will want to add the next option.

It could be nice if one could change the duration in URL. Of course this will get a bit tricky to select the correct option if user provides something that's not on the menu (you would need to dynamically add one more option as the currently selected one).

app/ports/views.py
```diff
@@ -151,18 +151,28 @@ def portdetail_build_information(request):
...
+    except ValueError:
+        return HttpResponse("'days' and 'days-ago' must be integer.")
     port_name = request.GET.get('port_name')
     port = Port.objects.get(name__iexact=port_name)
 
+    end_date = datetime.datetime.now(tz=datetime.timezone.utc) - datetime.timedelta(days_ago)
+    start_date = end_date - datetime.timedelta(days=days)
```
mojca: We have (or will have) the same functionality on `/statistics/`. It makes sense to write a function that will return two datetime objects from the given (`days`, `days_ago`).

app/ports/views.py
```diff
@@ -151,18 +151,28 @@ def portdetail_build_information(request):
...
-    submissions_last_30_days = Submission.objects.filter(timestamp__gte=datetime.datetime.now(tz=datetime.timezone.utc) - datetime.timedelta(days=30)).order_by('user', '-timestamp').distinct('user')
-    installations_last_30_days = PortInstallation.objects.filter(submission_id__in=Subquery(submissions_last_30_days.values('id')), port__iexact=port_name)
-    requested_count = installations_last_30_days.filter(requested=True).values('port').aggregate(Count('port'))
-    total_count = installations_last_30_days.values('port').aggregate(Count('port'))
-    version_distribution = installations_last_30_days.values('version').annotate(num=Count('version')).order_by('-num')
-    os_distribution_unsorted = list(installations_last_30_days.values('submission__os_version', 'submission__os_arch').annotate(num=Count('submission__user_id', distinct=True)).order_by('-num'))
+    submissions_last_x_days = Submission.objects.filter(timestamp__range=[start_date, end_date]).order_by('user', '-timestamp').distinct('user')
+    installations_last_x_days = PortInstallation.objects.filter(submission_id__in=Subquery(submissions_last_x_days.values('id')), port__iexact=port_name)
```
mojca: Inside this function you don't have any other submissions, right? Also, it's no longer true that this is "last X days". I would call this simply `submissions`, or maybe `submissions_in_timespaninterval/range` or `statistics_submissions[...]`. And the second one simply `port_installations`.

Out of curiosity: is this actually running the first query and transferring the results, or does this result in a nested SQL query / subquery?

app/ports/views.py
```diff
@@ -151,18 +151,28 @@ def portdetail_build_information(request):
...
+    requested_count = installations_last_x_days.filter(requested=True).values('port').aggregate(Count('port'))
```
mojca: It's ok as it is now, another name could be `requested_port_installations_count`.

app/ports/views.py
```diff
@@ -151,18 +151,28 @@ def portdetail_build_information(request):
...
+    os_distribution_unsorted = list(installations_last_x_days.values('submission__os_version', 'submission__os_arch').annotate(num=Count('submission__user_id', distinct=True)).order_by('-num'))
     os_distribution = sorted(os_distribution_unsorted, key=lambda x: (tuple(int(i) for i in x['submission__os_version'].split('.'))), reverse=True)
-    xcode_distribution_unsorted = list(installations_last_30_days.values('submission__xcode_version', 'submission__os_version').annotate(num=Count('submission__user_id', distinct=True)))
+    xcode_distribution_unsorted = list(installations_last_x_days.values('submission__xcode_version', 'submission__os_version').annotate(num=Count('submission__user_id', distinct=True)))
     xcode_distribution = sorted(xcode_distribution_unsorted, key=lambda x: (tuple(int(i) for i in x['submission__os_version'].split('.'))), reverse=True)
 
     installation_by_month = PortInstallation.objects.filter(port__iexact=port_name).annotate(month=TruncMonth('submission__timestamp')).values('month').annotate(num=Count('submission__user', distinct=True))[:12]
```
mojca: I'm confused. Won't this query also count installations from July 2018?

How could we make some tests for these queries?

app/ports/views.py
```diff
@@ -151,18 +151,28 @@ def portdetail_build_information(request):
...
-    installations_last_30_days = PortInstallation.objects.filter(submission_id__in=Subquery(submissions_last_30_days.values('id')), port__iexact=port_name)
-    requested_count = installations_last_30_days.filter(requested=True).values('port').aggregate(Count('port'))
-    total_count = installations_last_30_days.values('port').aggregate(Count('port'))
-    version_distribution = installations_last_30_days.values('version').annotate(num=Count('version')).order_by('-num')
-    os_distribution_unsorted = list(installations_last_30_days.values('submission__os_version', 'submission__os_arch').annotate(num=Count('submission__user_id', distinct=True)).order_by('-num'))
+    submissions_last_x_days = Submission.objects.filter(timestamp__range=[start_date, end_date]).order_by('user', '-timestamp').distinct('user')
+    installations_last_x_days = PortInstallation.objects.filter(submission_id__in=Subquery(submissions_last_x_days.values('id')), port__iexact=port_name)
+    requested_count = installations_last_x_days.filter(requested=True).values('port').aggregate(Count('port'))
+    total_count = installations_last_x_days.values('port').aggregate(Count('port'))
```
mojca: Similar here: `total_port_installations_count`

app/ports/views.py
```diff
@@ -151,18 +151,28 @@ def portdetail_build_information(request):
...
+    version_distribution = installations_last_x_days.values('version').annotate(num=Count('version')).order_by('-num')
```
mojca: Maybe `port_installations_by_port_version` (or some variant of that name).

app/ports/views.py
```diff
@@ -151,18 +151,28 @@ def portdetail_build_information(request):
...
+    os_distribution_unsorted = list(installations_last_x_days.values('submission__os_version', 'submission__os_arch').annotate(num=Count('submission__user_id', distinct=True)).order_by('-num'))
```
mojca: Maybe `port_installations_by_os_version_and_arch`.

app/ports/views.py
```diff
@@ -151,18 +151,28 @@ def portdetail_build_information(request):
...
     os_distribution = sorted(os_distribution_unsorted, key=lambda x: (tuple(int(i) for i in x['submission__os_version'].split('.'))), reverse=True)
-    xcode_distribution_unsorted = list(installations_last_30_days.values('submission__xcode_version', 'submission__os_version').annotate(num=Count('submission__user_id', distinct=True)))
+    xcode_distribution_unsorted = list(installations_last_x_days.values('submission__xcode_version', 'submission__os_version').annotate(num=Count('submission__user_id', distinct=True)))
```
mojca: Maybe `port_installations_by_os_and_xcode_version`

app/ports/views.py
arjunsalyan: > I'm confused. Won't this query also count installations from July 2018?

I had the same concern while using it, but it does not.

> How could we make some tests for these queries?

I am in the process of writing some fake submissions. I could think of only this, is there any other way?

app/ports/templates/ports/port-detail/installation_stats.html
arjunsalyan: Shall we restrict the user to use only some values (7, 30, 90, 180, 365)? We could then get rid of the repetition also.

app/ports/views.py
arjunsalyan: > Inside this function you don't have any other submissions, right? Also, it's no longer true that this is "last X days". I would call this simply `submissions`, or maybe `submissions_in_timespaninterval/range` or `statistics_submissions[...]`. And the second one simply `port_installations`.

Thank you very much.

> Out of curiosity: is this actually running the first query and transferring the results, or does this result in a nested SQL query / subquery?

As per the documentation, it should run a single query.

docs/API/v1.md
```diff
@@ -0,0 +1,84 @@
+## API V1
```
mojca: Can you please drop this commit?

app/ports/templates/ports/stats.html
```diff
@@ -182,7 +182,7 @@ <h6 class="text-center">Total Installations</h6>
                 </div>
                 <div class="col-lg-6 col-md-6 p-2">
                     <h6 class="text-center">Requested Installations</h6>
-                    <table class="table table-condensed border">
+                    <table class="table table-striped table-condensed border">
```
mojca: What does this do?

app/static/css/main.css
```diff
@@ -20,4 +20,19 @@ body{
     display: block;
     height: 0;
     overflow-y: hidden;
-}
\ No newline at end of file
+}
+
+.order_by_1{
+    background-color: #38610B;
```
mojca, line 26: We could do some shades of blue, but it's ok for now.

app/ports/views.py
```diff
@@ -148,27 +158,47 @@ def portdetail_build_information(request):
...
     port = Port.objects.get(name__iexact=port_name)
 
-    # Section for calculation of current stats
-    submissions_last_30_days = Submission.objects.filter(timestamp__gte=datetime.datetime.now(tz=datetime.timezone.utc) - datetime.timedelta(days=30)).order_by('user', '-timestamp').distinct('user')
-    installations_last_30_days = PortInstallation.objects.filter(submission_id__in=Subquery(submissions_last_30_days.values('id')), port__iexact=port_name)
-    requested_count = installations_last_30_days.filter(requested=True).values('port').aggregate(Count('port'))
-    total_count = installations_last_30_days.values('port').aggregate(Count('port'))
-    version_distribution = installations_last_30_days.values('version').annotate(num=Count('version')).order_by('-num')
-    os_distribution = installations_last_30_days.values('submission__os_version').annotate(num=Count('submission__os_version')).order_by('-num')
+    end_date = datetime.datetime.now(tz=datetime.timezone.utc) - datetime.timedelta(days_ago)
```
mojca: What is `datetime.timedelta(days_ago)`? What time unit?

app/ports/templates/ports/stats_port_installations.html
```diff
@@ -0,0 +1,73 @@
...
+    <input type="text" class="form-control" id="search-by" name="search-by" placeholder="Filter by port name">
+    <br>
+    <form class="form-inline">
+    <label class="form-inline" for="days-filter">Showing data for users who submitted stats in last: </label>
+    <select class="form-control form-inline ml-2" id="days-filter" name="days-filter">
+        <option value="7">7 days</option>
+        <option value="30" selected>30 days</option>
+        <option value="90">90 days</option>
+        <option value="180">180 days</option>
+        <option value="365">365 days</option>
```
mojca: You had some generic code for this elsewhere?

app/ports/templates/ports/stats.html
```diff
@@ -139,97 +112,62 @@
...
+    <br>
+    {% csrf_token %}
+    <form class="form-inline justify-content-center" method="get" action="{% url 'stats_home' %}#os-chart">
+        <label for="days"><h5>The charts below show stats for users who made submissions in last: </h5></label>
+        <select onchange="this.form.submit()" class="form-control ml-2" name="days" id="days">
+            <option class="format-option" value="7" >7 Days</option>
+            <option class="format-option" value="30" {% if days == 30 %}selected{% endif %}>30 Days</option>
+            <option class="format-option" value="90" {% if days == 90 %}selected{% endif %}>90 Days</option>
+            <option class="format-option" value="180" {% if days == 180 %}selected{% endif %}>180 Days</option>
+            <option class="format-option" value="365" {% if days == 365 %}selected{% endif %}>365 Days</option>
```
mojca: Isn't this code already generalized elsewhere?

mojca (commented):
Looks ok.

app/ports/views.py
```diff
@@ -151,18 +151,28 @@ def portdetail_build_information(request):
 
 
 def portdetail_stats(request):
+    days = request.GET.get('days', 30)
+    days_ago = request.GET.get('days_ago', 0)
+    try:
+        days = int(days)
+        days_ago = int(days_ago)
+    except ValueError:
+        return HttpResponse("'days' and 'days-ago' must be integer.")
```
umeshksingla: `days-ago` -> `days_ago`

app/ports/views.py
arjunsalyan: I believe the proper syntax would be:
```python
datetime.timedelta(days=days_ago)
```

docs/API/v1.md
arjunsalyan: Thanks, dropped.

app/ports/templates/ports/stats.html
arjunsalyan: We received a feedback about using tables with rows having alternate colours. `table-striped` does that. But these tables were removed altogether, later.

app/ports/templates/ports/stats_port_installations.html
arjunsalyan: Thank you, I have removed the repetition.

app/ports/templates/ports/stats.html
arjunsalyan: Yes, thanks.

app/ports/templates/ports/stats.html
mojca: Oh, sorry, thanks. I was confusing striped vs. stripped.

app/ports/views.py
umeshksingla: Should we not do this query over either `requested_port_installations_count` or `total_port_installations_count`?
Same for the os_version and os_arch and xcode_version queries below.

arjunsalyan:
> We can probably remove the `User` model now?

Yes, I will create a backup (we won't need it, but in case ...) and then remove `User` and `OSDistribution` models.

app/ports/migrations/0004_add-fields-to-stats.py
```diff
@@ -0,0 +1,41 @@
...
+    operations = [
+        migrations.AddField(
+            model_name='submission',
+            name='build_arch',
+            field=models.CharField(default='', max_length=20),
+        ),
+        migrations.AddField(
+            model_name='submission',
+            name='platform',
+            field=models.CharField(default='', max_length=20),
```
mojca, line 22: I'm not 100% sure, but we could actually use `darwin` as default here :)

Similar question as above: could the migration automatically set `darwin`; this time simply to all existing entries?

app/ports/models/stats.py
```diff
@@ -39,9 +42,12 @@ class Meta:
             models.Index(fields=['user', '-timestamp']),
             models.Index(fields=['user', 'timestamp']),
             models.Index(fields=['os_version']),
+            models.Index(fields=['os_version', 'os_arch']),
```
mojca, line 28: Doesn't this need to be `os_version, build_arch`?

app/ports/migrations/0004_add-fields-to-stats.py
```diff
@@ -0,0 +1,41 @@
...
+
+    dependencies = [
+        ('ports', '0003_add-field-cxx_stdlib'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='submission',
+            name='build_arch',
+            field=models.CharField(default='', max_length=20),
```
mojca, line 17: Can the migration file contain something like the following (pseudocode):
* for all existing entries where os_version in (10.7, 10.8, 10.9, ... 10.14) set build_arch = x86_64
* for all existing entries where os_arch = ppc, set build_arch = ppc
* for all existing entries where os_version = 10.5 and os_arch = i386, set build_arch = i386

app/ports/models/stats.py
```diff
@@ -28,8 +28,11 @@ class Submission(models.Model):
     os_version = models.CharField(max_length=10)
     xcode_version = models.CharField(max_length=10)
     os_arch = models.CharField(max_length=20)
+    build_arch = models.CharField(max_length=20, default='')
+    platform = models.CharField(max_length=20, default='')
     macports_version = models.CharField(max_length=10)
     cxx_stdlib = models.CharField(max_length=20, default='libc++')
+    raw_json = JSONField(default=dict)
```
mojca, line 18: Cool. I checked and django uses binary json by default.

app/ports/models/stats.py
mojca: Thinking a bit further. Will we need to index on `['platform', 'os_version', 'build_arch']`? As soon as someone submits statistics from Linux, we might need to adapt the code in the queries, and then we'll probably need to check all three rather than just two..

app/ports/templates/ports/port-detail/installation_stats.html
```diff
@@ -186,6 +189,32 @@ <h4 class="modal-title">How are monthly installations calculated?</h4>
       }
     </script>
 
+
+    <!-- OS VERSIONS AND ARCHITECTURES -->
+    <script type="text/javascript">
+        google.charts.load('current', {'packages':['corechart']});
+        google.charts.setOnLoadCallback(drawChart);
+        function drawChart() {
+            {% regroup port_installations_by_stdlib_and_os_arch|dictsort:"submission__cxx_stdlib" by submission__cxx_stdlib as stdlib_chart_stdlib_parent %}
```
mojca: We could make a backup of this commit and potentially even remove it, and implement it the way we discussed today (it doesn't need to be done immediately).

app/ports/models/stats.py
umeshksingla: True. Otherwise, how would we accommodate `os_version` submitted from Linux? I think filtering by the `os_platform` is necessary until we support Linux submissions in our queries.

app/ports/models/stats.py
```diff
@@ -28,7 +11,11 @@ class Submission(models.Model):
     os_version = models.CharField(max_length=10)
     xcode_version = models.CharField(max_length=10)
     os_arch = models.CharField(max_length=20)
+    build_arch = models.CharField(max_length=20, default='')
+    platform = models.CharField(max_length=20, default='')
```
umeshksingla, line 15: It would be better if you keep it `os_platform` consistent with the received JSON.

app/ports/templates/ports/index.html
```diff
@@ -8,7 +8,7 @@
         {% csrf_token %}
         <div class="form-group mb-0">
             <label class="radio-inline mr-3"><input onclick="switch_search()" type="radio" name="search_by" value="name" checked>Search by port-name</label>
-            <label class="radio-inline"><input onclick="switch_search()" type="radio" name="search_by" value="description">Search by Description</label>
+            <label class="radio-inline"><input onclick="switch_search()" type="radio" name="search_by" value="description">Search by description</label>
```
umeshksingla, line 11: A very small issue but I see the text "Search by ..." is not aligned with the radio button and also, is very close to it.

app/ports/templates/ports/port-detail/installation_stats.html
```diff
@@ -85,133 +111,169 @@ <h4 class="modal-title">How are monthly installations calculated?</h4>
...
+    </script>
 
-      // Set a callback to run when the Google Visualization API is loaded.
-      google.charts.setOnLoadCallback(drawChart);
 
-      // Callback that creates and populates a data table,
-      // instantiates the pie chart, passes in the data and
-      // draws it.
-      function drawChart() {
+    <!-- OS VERSIONS AND ARCHITECTURES -->
```
umeshksingla: Is it just for me or the value `libstdc++` breaks into two in the legend for others too?
Screen Shot 2019-07-28 at 1 00 51 PM

Also, if we plan to keep this part or comment it, `<!-- OS VERSIONS AND ARCHITECTURES -->` would need to be changed to avoid confusion.

app/ports/migrations/0004_add-fields-to-stats.py
arjunsalyan: Do we need to supply such code with the repo? I was planning to do this, but not commit the code.

app/ports/models/stats.py
arjunsalyan: Thank you. We will need an index on all three: `['platform', 'os_version', 'build_arch']`.

app/ports/templates/ports/stats_port_installations.html
```diff
@@ -0,0 +1,71 @@
...
+        <strong>Total</strong>
+        <button id="total_count" onclick="sort(this)" class="btn btn-sm ml-3">&#9650;</button><button id="-total_count" onclick="sort(this)" class="btn btn-sm">&#9660;</button>
+    </div>
+    <div class="col-3 pr-4 border-left border-top border-bottom border-right text-center">
+        <strong>Requested</strong>
+        <button id="req_count" onclick="sort(this)" class="btn btn-sm ml-3">&#9650;</button><button id="-req_count" onclick="sort(this)" class="btn btn-sm">&#9660;</button>
+    </div>
+</div>
+<div id="loading-image" class="container text-center" style="height: 500px; display: none">
+    <img src="{% static 'images/giphy.gif' %}">
```
umeshksingla, line 54: A very stupid thing to ask but can we just rename it to `loader.gif` or something?

app/ports/views.py
```diff
@@ -225,13 +264,13 @@ def stats(request):
...
+    submissions_unique = Submission.objects.filter(id__in=Subquery(submissions_last_x_days.values('id')))
     macports_distribution = submissions_unique.values('macports_version').annotate(num=Count('macports_version'))
-    os_distribution = submissions_unique.values('os_version').annotate(num=Count('os_version'))
-    xcode_distribution = submissions_unique.values('xcode_version').annotate(num=Count('xcode_version'))
-    port_distribution = PortInstallation.objects.filter(submission_id__in=Subquery(submissions_last_30_days.values('id'))).values('port').annotate(num=Count('port')).order_by('-num').exclude(port__iexact='mpstats-gsoc')[:50]
-    req_port_distribution = PortInstallation.objects.filter(submission_id__in=Subquery(submissions_last_30_days.values('id')), requested=True).values('port').annotate(num=Count('port')).order_by('-num')[:50]
+    os_distribution_unsorted = list(submissions_unique.values('os_version', 'os_arch').annotate(num=Count('user_id', distinct=True)))
+    os_distribution = sorted(os_distribution_unsorted, key=lambda x: int(x['os_version'].replace(".", '')), reverse=True)
```
umeshksingla, line 268: @mojca can you verify this one? It seems to me that if it's something like 10.13.1 vs 10.14, the ordering will not be right.
I was thinking we will do the splitting part while rendering the chart to keep the order of OS versions consistent. Also, a separate function to sort the way we intend to would be nice and will make it clearer with an example documented in its docstring.

app/ports/views.py
```diff
@@ -225,13 +264,13 @@ def stats(request):
...
+    xcode_distribution_unsorted = list(submissions_unique.values('xcode_version', 'os_version').annotate(num=Count('user_id', distinct=True)))
+    xcode_distribution = sorted(xcode_distribution_unsorted, key=lambda x: (tuple(int(i) for i in x['os_version'].split('.'))), reverse=True)
```
umeshksingla, line 270: Same here.

app/ports/migrations/0004_add-fields-to-stats.py
mojca: I'm not sure. In my understanding one of the ideas of migrations is to do precisely that: add missing tables, columns, and populate them with some values. But then again, since we have only been collecting statistics at a single site, there will be precisely one user of that migration anyway, so it boils down how you want to do it.

(The same is true for all migrations mostly, and the same argument applies that we hardly need migrations during the development period.)

In principle I'm also fine with manually running the changes. You had 160 submissions so far. I would personally probably not be manually editing each one of them, and if you run slightly more complex queries, it doesn't make much of a difference whether you just execute them, or store them first, then run migrations to auto-populate everything.

I leave it up to you to decide.

@umeshksingla?

app/ports/views.py
```diff
@@ -240,10 +279,69 @@ def stats(request):
...
         'macports_distribution': macports_distribution,
         'xcode_distribution': xcode_distribution,
-        'req_port_distribution': req_port_distribution
+        'days': int(days),
+        'allowed_days': allowed_days
+    })
+
+
+def stats_port_installations(request):
+    days = int(request.GET.get('days', 30))
```
umeshksingla: This can potentially fail.

app/ports/views.py
```diff
@@ -240,10 +279,69 @@ def stats(request):
...
+        'days': days,
+        'first': first,
+        'second': second,
+        'third': third,
+        'allowed_days': allowed_days
+    })
+
+
+def stats_port_installations_filter(request):
+    days = int(request.GET.get('days', 30))
```
umeshksingla: This needs a check again. Actually, it'd be better if the argument validation is consistent across the code. You'll not need to work on how to add it again and again, and also will help you remember.

app/ports/views.py
```diff
@@ -240,10 +279,69 @@ def stats(request):
...
+    order_by_1 = str(request.GET.get('order_by_1', '-total_count'))
+    order_by_2 = str(request.GET.get('order_by_2', '-req_count'))
+    order_by_3 = str(request.GET.get('order_by_3', 'port'))
+    search_by = str(request.GET.get('search_by', ''))
+
+    submissions_unique = Submission.objects.filter(timestamp__gte=datetime.datetime.now(tz=datetime.timezone.utc)-datetime.timedelta(days=days)).order_by('user', '-timestamp').distinct('user')
+    installations = PortInstallation.objects.order_by('port')\
```
umeshksingla, line 341: I'll be okay if you would like to break this query into multiple with appropriate variable names and comments.

app/static/js/ajax-port-installations.js
```diff
@@ -0,0 +1,109 @@
...
+
+$(function () {
+    $('#table-header').ready(function () {
+        var state = getPageState();
+        runPortsSortAjax(1, state.order_by_1, state.order_by_2, state.order_by_3, state.days, state.search_by);
+    });
+});
+
+
+function sort(e) {
```
umeshksingla, line 63: I will check on this but is there no way to doing this some other way? maybe a Django package? I am sure no one would like to modify this later if there's a need.

app/ports/migrations/0004_add-fields-to-stats.py
arjunsalyan: > I'm not 100% sure, but we could actually use `darwin` as default here :)
> 
> Similar question as above: could the migration automatically set `darwin`; this time simply to all existing entries?

If we had set `default=darwin`, then this migration would automatically set `darwin` in existing entries.

I would add `darwin` as default, thanks.

app/ports/views.py
```diff
@@ -240,10 +279,69 @@ def stats(request):
...
+    values = [first, second, third]
+    acceptable_values = ['port', '-port', 'total_count', '-total_count', '-req_count', 'req_count']
+    allowed_days = [7, 30, 90, 180, 365]
+
+    if days not in allowed_days:
+        return HttpResponse("'days' key must have an integer value from the list: {}".format(allowed_days))
+
+    for i in values:
+        if i not in acceptable_values:
+            return HttpResponse("'{}' is an invalid value to sort. Allowed values: {}".format(acceptable_values))
```
umeshksingla: I think you missed to pass `i` in the `.format()`.

app/ports/models/stats.py
arjunsalyan: And also `['os_version', 'build_arch', 'cxx_stdlib']`.

app/ports/views.py
```diff
@@ -148,27 +158,49 @@ def portdetail_build_information(request):
...
-    version_distribution = installations_last_30_days.values('version').annotate(num=Count('version')).order_by('-num')
-    os_distribution = installations_last_30_days.values('submission__os_version').annotate(num=Count('submission__os_version')).order_by('-num')
+    end_date = datetime.datetime.now(tz=datetime.timezone.utc) - datetime.timedelta(days=days_ago)
+    start_date = end_date - datetime.timedelta(days=days)
 
-    installation_by_month = PortInstallation.objects.filter(port__iexact=port_name).annotate(month=TruncMonth('submission__timestamp')).values('month').annotate(num=Count('submission__user', distinct=True))[:12]
-    version_by_month = PortInstallation.objects.filter(port__iexact=port_name).annotate(month=TruncMonth('submission__timestamp')).values('month', 'version').annotate(num=Count('submission__user', distinct=True))[:12]
+    # Section for calculation of current stats
+    submissions = Submission.objects.filter(timestamp__range=[start_date, end_date]).order_by('user', '-timestamp').distinct('user')
+    port_installations = PortInstallation.objects.filter(submission_id__in=Subquery(submissions.values('id')), port__iexact=port_name)
```
umeshksingla, line 179: I think it has been mentioned somewhere before, that a lot of code in `portdetail_stats` and `stats` is repeated and can use some common methods.

app/ports/migrations/0004_add-fields-to-stats.py
umeshksingla: I'd just run the django migrations even if it's adding the default value to all rows. Because we have to add it to the model anyway and eventually migration files so it'll try to overwrite them.

app/ports/models/stats.py
arjunsalyan: Okay, thanks.

app/ports/templates/ports/stats_port_installations.html
arjunsalyan: Yes, thank you. I will do this.

app/ports/views.py
arjunsalyan: For Xcode, 10.2.1 is placed below 10.3. So, I believe this shall work. I will test it out.

A lot of refactoring can be done in the stats code. I will do this, thanks.

app/ports/views.py
arjunsalyan: Oh, I missed this one. Thanks

app/static/js/ajax-port-installations.js
arjunsalyan: I searched for other ways of doing this. But none was a good way to handle large number of ports that we have. There are also packages in JQuery, but again they will first load all the objects and then perform the operations.

There is no package to achieve efficiently the suggestion mentioned here: https://github.com/macports/macports-webapp/issues/39, so I went with writing custom code.

Presence of ajax makes it even more hard to follow.

app/ports/views.py
arjunsalyan: Oh, thank you very much.

app/ports/views.py
arjunsalyan: Probably I can write a separate file something like `validator.py`. What do you suggest?

arjunsalyan: I have written all the validators inside a separate file.

app/ports/views.py
mojca: Can you break a query without actually fetching the results from the database and just assign variables?

app/ports/views.py
```diff
@@ -246,12 +246,11 @@ def all_builds_filter(request):
...
-    try:
-        if int(days) not in allowed_days:
-            return HttpResponse("'days' received an invalid value.")
-    except ValueError:
-            return HttpResponse("'days' must be an integer.")
+
+    # Validate days
+    check, message = validate_stats_days(days)
+    if check is False:
+        return HttpResponse(message)
```
mojca, line 253: One option would be to simply round to the nearest "valid" value instead, and maybe drop some invalid arguments instead of showing the error, but let's not touch this now..

PR #45, opened 2019-07-20
Add API Documentation

arjunsalyan:


arjunsalyan:
I am not very sure about the format of this file. Could we please first verify if the attempted format is fine, and discuss any changes in it?

docs/API-V1/1-INTRODUCTION.md
```diff
@@ -0,0 +1,9 @@
+## API V1
+
+The API v1 of the webapp is callable at /api/v1/{call}?{params}. The response is encoded as JSON object(s).
+
+#### Index
+
+- **SINGLE-PORT**: Information, build history and installation stats related to a particular port.
```
mojca: These could be relative (clickable) links.

docs/API-V1/2-SINGLE-PORT.md
```diff
@@ -0,0 +1,45 @@
...
+
+#### `- GET /port/`
+
+Fetch the general information about a port i.e. the data extracted from the `Portfile`.
+
+Parameters:
+
+| key | type | details | default |
+|------|-----|----|----|
+| name | str, required | Name of the port | -- |
```
mojca: I would use `/port/<name>/` here to get "full info" about port

docs/API-V1/2-SINGLE-PORT.md
```diff
@@ -0,0 +1,45 @@
...
+| field | str, optional | Select only a particular field. If not provided, then all fields are sent.| All |
+
+#### - `GET /port/builds/`
```
mojca: `/port/<name>/builds/`

mojca (commented):
I've seen that [buildbot](https://github.com/buildbot/buildbot/tree/master/master/buildbot/spec/types) uses [raml](https://raml.org/).

Maybe that's an interesting option for proper specification.
But I don't want to enforce anything. I don't expect too many options to be supported, and the main point at the moment is to write specification, which from my point of view could be on a piece of paper, as long as it does the job.

docs/API-V1/1-INTRODUCTION.md
arjunsalyan: Thank you

docs/API-V1/1-INTRODUCTION.md
```diff
@@ -0,0 +1,10 @@
+## API V1
+
+The API v1 of the webapp is callable at /api/v1/{call}?{params}. The response is encoded as JSON object(s).
```
mojca: Is the first sentence still correct? I find it at least a bit misleading.

docs/API-V1/2-SINGLE-PORT.md
```diff
@@ -0,0 +1,42 @@
...
+
+Fetch the general information about a port i.e. the data extracted from the `Portfile`.
+
+Parameters:
+
+| key | type | details | default |
+|------|-----|----|----|
+| field | str, optional | Select only a particular field. If not provided, then all fields are sent.| All |
+
+#### - `GET /port/builds/{port-name}`
```
mojca: What if we start shipping a port called `builds`? We should probably switch to `/ports/` for general information about builds.

Then again, if we are talking about builds of one particular port, we could use `/port/<name>/builds`, for example.

If we use `/port/<name>/builds` for this, we could use `/port/<name>/info` for general port information.

What I miss here is port health status, either total, or for a particular builder.

docs/API-V1/2-SINGLE-PORT.md
```diff
@@ -0,0 +1,42 @@
+## Details of a Single Port
```
mojca, line 1: What I generally miss in this document (to aid the discussion & development) is an example of a response.
(We could potentially use some real data for some common port that's simple enough, yet complex enough to demonstrate the capabilities.)

docs/API-V1/3-FILTER-PORTS.md
```diff
@@ -0,0 +1,19 @@
+## Filter Ports
+
+#### `- GET /ports/`
```
mojca, line 3: I miss the part that just gives you the count of ports, and a way to limit the results that you fetch. (If one would want to create some UI with pagination, it would be nice to just get the info for the 20 desired ports.)

docs/API-V1/2-SINGLE-PORT.md
```diff
@@ -0,0 +1,42 @@
...
+#### - `GET /port/builds/{port-name}`
+
+Fetch build history of a given port.
+
+Parameters
+
+| key | type | details | default |
+|------|-----|----|----|
+| count | int, optional | Number of builds to be fetched for each builder. | All |
+| builder | str, optional | Specify one more builders, separated by comma (`,`). | All |
```
mojca: one or more

docs/API-V1/2-SINGLE-PORT.md
```diff
@@ -0,0 +1,42 @@
...
+
+Fetch installation statistics of a given port.
+
+Parameters
+
+| key | type | details | default
+|------|-----|----|----|
+| days_ago | int, optional | Number of days behind current day, to calculate stats. | 0 |
+| days | int, optional | Number of days behind `days_ago`. Only the submissions made in this period are used to generate the stats. | 30 |
+| criteria | optional | Comma separated list of the criterial of stats. Options: ['total_count', 'req_count', 'os', 'xcode', 'os_arch', 'cxx_stdlib'] | All |
```
mojca: The first two are clear. For `criteria` it's not entirely clear to me what one would pass to the function and what one would get back. Also `os_arch` may need to be replaced.

docs/API-V1/2-SINGLE-PORT.md
```diff
@@ -0,0 +1,65 @@
...
+
+Fetch build history of a given port.
+
+Parameters
+
+| key | type | details | default |
+|------|-----|----|----|
+| count | int, optional | Number of builds to be fetched for each builder. | All |
+| builder | str, optional | Filter the builds by builder | All |
+| status | str, optional | Filter by the status of builds | All |
```
mojca, line 23: It's not clear to me what the `status` might be in this case (I may know it personally, but it's not entirely clear to others). I would somehow prefer to offer the official numbers for Success Failure, but then we could not distinguish failed builds from failed dependency installations. Even so I'm not sure what the best parameter to status could or should be.

docs/API-V1/2-SINGLE-PORT.md
```diff
@@ -0,0 +1,65 @@
...
+| builders | str, optional | Comma separated list of builders | All |
+
+```json
+    [
+        {
+            "builder1": "build successful",
+            "builder2": "failed install-port",
+        }
+    ]
+```
```
mojca, line 65: Such examples would be very helpful everywhere (for all api calls), with both sample request + sample answer. And in this case maybe a real builder name as an example.

mojca (commented):
I need more time, I'm just leaving a few initial notes.

arjunsalyan:
This old and incomplete API is now obsolete and has been taken care of separately.

Issue #29, opened 2019-07-09
Is django needed both outside and inside the requirements file to Travis?


umeshksingla:
Thread: https://github.com/macports/macports-webapp/pull/13#discussion_r294192381

There's `Django==2.1.7` in requirements.txt which probably needs to be removed.

_Originally posted by @umeshksingla in https://github.com/macports/macports-webapp/pull/13_

arjunsalyan:
@mojca previously suggested that we might want to test only for the latest versions.

Removing `Django` from `requirements.txt` would make us to install `django` separately (separately from `pip install -r requirements.txt`), and this is the only reason I am not sure about this. We may either remove the version and just let Travis test for the latest version, or may ourselves add some latest version. I have no clue about how to best deal with this?

mojca:
No, I didn't want to suggest that we need to test **exclusively** the latest version. It doesn't hurt to have multiple tests setup, even if there's probably not much of a difference (I don't care that much if we have just one version tested). I would not remove django from that list, I would just probably remove the old hardcoded version.

It's two-fold problem. Either we put all the versions there and ensure that it will run forever, but we would eventually forget to update them (I noticed that @rajdeepbharati set up some bot to update dependencies versions in some of his repositories), or we risk that a newer version of one dependency will break the app, at which point we'll need to restrict it further later.

arjunsalyan:
We are no longer testing with multiple Django versions, but just with the fixed version from requirements.txt. So I guess this issue is not relevant now.

PR #55, opened 2019-07-29
requirements.txt: Fix requests/urllib3 incompatibility

neverpanic:
ERROR: requests 2.21.0 has requirement urllib3<1.25,>=1.21.1, but you'll have urllib3 1.25.3 which is incompatible.

### DIFF
--- a/app/requirements.txt
+++ b/app/requirements.txt
@@ -36,7 +36,7 @@ python-coveralls==2.9.1
 python-dateutil==2.8.0
 pytz==2018.9
 PyYAML==5.1.1
-requests==2.21.0
+requests==2.22.0
 Send2Trash==1.5.0
 six==1.12.0
 soupsieve==1.9


mojca:
@arjunsalyan: There is also a (harmless) warning
```
WARNING: You are using pip version 19.1.1, however version 19.2.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
```

arjunsalyan:
> @arjunsalyan: There is also a (harmless) warning
> 
> ```
> WARNING: You are using pip version 19.1.1, however version 19.2.1 is available.
> You should consider upgrading via the 'pip install --upgrade pip' command.
> ```

Maybe now is time that we also address https://github.com/macports/macports-webapp/issues/29

We probably decided to keep some upper limit for most of the requirement-versions, and for some like `pip` itself it might be better to let is install the most recent version?

PR #56, opened 2019-07-29
Add Tests for Statistics

arjunsalyan:
Tests for all the the functionality related to statistics written so far will be covered here.

### DIFF
--- a/app/ports/tests/sample_data/submissions.json
+++ b/app/ports/tests/sample_data/submissions.json
@@ -90,12 +90,12 @@
                 "version": "1.2"
             },
             {
-                "name": "port-D1",
+                "name": "port-A4",
                 "version": "1.0",
                 "requested": "true"
             },
             {
-                "name": "port-D2",
+                "name": "port-A3-diff",
                 "version": "1.12",
                 "requested": "true"
             }
@@ -125,12 +125,12 @@
                 "requested": "true"
             },
             {
-                "name": "port-D1",
+                "name": "port-A4",
                 "version": "1.0",
                 "requested": "true"
             },
             {
-                "name": "port-D2",
+                "name": "port-A3-diff",
                 "version": "1.12"
             }
         ]
@@ -159,12 +159,12 @@
                 "requested": "true"
             },
             {
-                "name": "port-D1",
+                "name": "port-A4",
                 "version": "1.0",
                 "requested": "true"
             },
             {
-                "name": "port-D2",
+                "name": "port-A3-diff",
                 "version": "1.12"
             }
         ]
@@ -193,12 +193,12 @@
                 "requested": "true"
             },
             {
-                "name": "port-D1",
+                "name": "port-A2",
                 "version": "1.0",
                 "requested": "true"
             },
             {
-                "name": "port-D2",
+                "name": "port-A3-diff",
                 "version": "1.12"
             }
         ]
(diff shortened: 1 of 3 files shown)


app/ports/tests/sample_data/submissions.json
```diff
@@ -0,0 +1,206 @@
...
+        "os": {
+            "macports_version": "2.5.4",
+            "osx_version": "10.14",
+            "os_arch": "i386",
+            "os_platform": "darwin",
+            "cxx_stdlib": "libc++",
+            "build_arch": "x86_64",
+            "gcc_version": "none",
+            "prefix": "/opt/local",
+            "xcode_version": "10.3"
```
mojca: We could support `clt_version` as well, but then again, we need tests for when the value is present and when it is not, so that's fine.

app/ports/tests/test_statistics.py
```diff
@@ -0,0 +1,61 @@
...
+class TestStatistics(TestCase):
+    def setUp(self):
+        self.client = Client()
+
+    @classmethod
+    def setUpTestData(cls):
+        Port.load(TEST_PORTINDEX_JSON)
+
+    def test_submission(self):
+        submission_body = 'submission[data]={\
```
mojca: I would probably use `"""` instead of having to escape everything.

app/ports/tests/sample_data/submissions.json
```diff
@@ -0,0 +1,206 @@
+[
+    {
+        "id": "XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXX1",
+        "os": {
+            "macports_version": "2.5.4",
+            "osx_version": "10.14",
+            "os_arch": "i386",
+            "os_platform": "darwin",
+            "cxx_stdlib": "libc++",
```
umeshksingla, line 9: I do not get `cxx_stdlib` value on running `mpstats show` or `mpstats-gsoc show`. Is it okay to set `libc++` as default?

umeshksingla: Sorry, I didn't have the latest version.

mojca: Other users who don't have the latest version may submit statistics this way. What I would do is check whether platform is `darwin` and `osx_version` is at least `10.9`, and set `libc++` in that case. If `osx_version` is smaller than that, I wouldn't dare betting about the value, and well ... we can simply leave it empty.

Maybe one of the unit tests could also cover submission from linux?

app/ports/tests/test_statistics.py
arjunsalyan: Thank you.

app/ports/tests/sample_data/submissions.json
arjunsalyan: Oh, I just realised that users who do not have the latest version might not be able to submit stats, and the code would raise `KeyError`. I should make use of `.get`.

> Other users who don't have the latest version may submit statistics this way. What I would do is check whether platform is `darwin` and `osx_version` is at least `10.9`, and set `libc++` in that case. If `osx_version` is smaller than that, I wouldn't dare betting about the value, and well ... we can simply leave it empty.

For submissions, I did not want to set any defaults. But I am probably less knowledgeable with `stdlib` and similar configs, so I would go with this.

mojca: Thanks for fixing the reading of data.
Yes, defaults can be left empty (except for `stdlib` in which case it's possible to know the value for 10.9+).

arjunsalyan:
A completely new model to handle statistics and related tests have been added. This PR is now obsolete.

PR #57, opened 2019-07-29
docker: Support running as unprivileged user

neverpanic:
For our production deployment, I do not want to run the container with elevated privileges. Instead, allow specifying a user a build argument, add the required chown statements for paths that should be writable during production and switch the nginx port to a port from the unprivileged port range so that non-root users can bind to it.

This allows successfully running the container using the following command line:

```sh
docker run -d \
    --publish 8080:8080 \
    --user=$uid:$gid \
    --env-file=env \
    -v /postgres/socket/dir:/postgres/socket/dir \
    -v /etc/passwd:/etc/passwd:ro \
    --name=macports-webapp macports-webapp
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
 FROM ubuntu:18.04
+ARG USER=0:0
 
 # Install required packages and remove the apt packages cache when done.
 
@@ -31,9 +32,6 @@ COPY config/supervisor.conf /etc/supervisor/conf.d/
 COPY app/requirements.txt /code/app/
 RUN pip3 install -r /code/app/requirements.txt
 
-RUN touch /var/log/buildhistorycron.log
-RUN touch /var/log/portinfocron.log
-
 # Setup cron
 COPY config/crons /etc/cron.d/crons
 RUN chmod 0644 /etc/cron.d/crons
@@ -41,5 +39,11 @@ RUN chmod 0644 /etc/cron.d/crons
 # add (the rest of) our code
 COPY . /code/
 
-EXPOSE 80
+RUN touch /var/log/buildhistorycron.log
+RUN touch /var/log/portinfocron.log
+RUN chown -R ${USER} /var/log/buildhistorycron.log /var/log/portinfocron.log /var/log/supervisor /var/log/nginx /var/lib/nginx /code
+RUN chown ${USER} /run
+
+USER ${USER}
+EXPOSE 8080
 CMD ["/usr/bin/supervisord", "-n"]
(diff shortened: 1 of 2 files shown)


arjunsalyan:
I have a concern about the cron jobs. To prevent them from running in a bare environment, I first export all variables to a file and then load them using `BASH_ENV` in the `config/crons` file.

But after removing `DB_PASSWORD` from the env file, how will the app and crons access the password? The app requires an env variable `DB_PASSWORD`. Sorry, if I am totally off track here.

neverpanic:
That's not a problem in our deployment, because it is totally password-less. Authentication against the database happens through the database server checking the UID of the process that is connected to the unix socket over which client and server talk. The only thing to consider is whether the cron job runs with the right user.

arjunsalyan:
I am getting this error while running the image:
```
The path /postgres/socket/dir
is not shared from OS X and is not known to Docker.
```
I tried to add `/postgres` to shared paths without any success.

**UPDATE**: Using the earlier `env-file` method worked fine.

**UPDATE 2**: The crons are working fine (But I used `DB_PASSWORD` inside env file).

arjunsalyan:
> The only thing to consider is whether the cron job runs with the right user.

Currently, they are running as `root`.

neverpanic:
`/postgres/socket/dir` needs to be replaced with whatever directory PostgreSQL puts its socket in – which is probably specific to our deployment.

Unfortunately when building the container with a user != root, cron doesn't start. I may end up just deploying the cron mechanism outside of the container and do the equivalent of `docker exec macports-webapp $cronjob`.

arjunsalyan:
On Ubuntu based EC2 instance, the cron jobs are working for me.
`docker exec macports-webapp $cronjob` seems to be another solution.

We run two commands using crons:

1. Fetch the build history every hour:
```
python3 /code/app/manage.py fetch-build-history
```

2. Update the ports information, can also run every hour.
```
python3 /code/app/manage.py update-portinfo
```

neverpanic:
I'm assuming on your EC2 instance this works fine because you haven't provided a `USER` build argument when building the docker container, which means it will default to root.

arjunsalyan:
I used the following command to start the container:

```
sudo docker run -d \
    --publish 80:8080 \
    --user=$uid:$gid \
    --env-file=env \
    --name=macports-webapp macports-webapp
```
with all environment variables in the `env` file.

Probably it is because of `sudo`?

arjunsalyan:
Oh, I am sorry. I probably missed the argument that needs to be provided while building the image.

PR #58, opened 2019-07-30
Improve Submission Model

arjunsalyan:
I have made the change to use `.get()` for keys that might be missing.

Regarding defaults, shall we discuss more before I implement?


mojca:
This commit looks good as it is, so I'll simply merge it now, and defaults can be added in a later commit.

I don't see any reasonable defaults, except for:
* making the platform `darwin` (not that it will make any difference)
* if macOS version exists and is at least `10.9`, make `libc++` default.

Please also add support for `clt_version` soon.

PR #60, opened 2019-07-31
Port Statistics: UI Improvements

arjunsalyan:


app/ports/templates/ports/port-detail/installation_stats.html
```diff
@@ -18,8 +20,12 @@
     </select>
 </form>
 <br>
+<h6 class="text-center">
+    <span class="text-muted">Duration selected:</span> <u>{{ start_date|date:"d M Y" }}</u> <span class="text-muted">to</span> <u>{{ end_date|date:"d M Y" }}</u>
```
mojca, line 24: Could we perhaps use YYYY-MM-DD here (that's more "international" than the US vs. European dates with inconsistent order of days & months)?

app/ports/views.py
```diff
@@ -196,8 +196,10 @@ def portdetail_stats(request):
         'port_installations_by_month': port_installations_by_month,
         'port_installations_by_version_and_month': port_installations_by_version_and_month,
         'port_installations_by_os_stdlib_build_arch': port_installations_by_os_stdlib_build_arch,
-        'days': days,
-        'days_ago': days_ago,
+        'days': int(days),
+        'days_ago': int(days_ago),
```
mojca: What happens when `days` is equal to `'a'`?

A simple test suggests says that the page doesn't load at all.

We could probably ignore invalid arguments and set them to default values, but I'm not 100% sure about the best approach, and I'm open to other suggestions.

app/ports/templates/ports/port-detail/installation_stats.html
```diff
@@ -90,7 +100,7 @@ <h4 class="modal-title">How are monthly installations calculated?</h4>
                 <button type="button" class="close" data-dismiss="modal">&times;</button>
             </div>
             <div class="modal-body">
-                <p>For each month, a user is considered to have the port installed, if in anyone of the submissions
+                <p>For each month, a user is considered to have the port installed, if in any one of the submissions
                     made by the user in that month- the port is present.</p>
```
mojca, line 104: One should not use `-` without the space in front of it (unless it's a compound world), and usually it should be `endash` or `emdash`.

But maybe this could be reworded anyway.

app/ports/views.py
```diff
@@ -196,8 +196,10 @@ def portdetail_stats(request):
         'port_installations_by_month': port_installations_by_month,
         'port_installations_by_version_and_month': port_installations_by_version_and_month,
         'port_installations_by_os_stdlib_build_arch': port_installations_by_os_stdlib_build_arch,
-        'days': days,
-        'days_ago': days_ago,
+        'days': int(days),
```
umeshksingla: This is okay but rather than doing this conversion at multiple places in the function, I'd just do once at the top.

umeshksingla (approved):

app/ports/views.py
arjunsalyan: We print a message saying that `'a'` is an invalid value and it must be an integer.

We can ignore and use some defaults. I will change this.

app/ports/templates/ports/port-detail/installation_stats.html
arjunsalyan: Thank you. It definitely can be reworded.

app/ports/templates/ports/port-detail/installation_stats.html
arjunsalyan: Okay, thanks. But maybe we could use the three character naming of months, it just feels easier to grasp?

PR #61, opened 2019-07-31
Workspace (only for stats)

arjunsalyan:


app/ports/utilities/sort_by_version.py
```diff
@@ -0,0 +1,6 @@
+def sort_dicts_by_version(dictionary, key):
+    return sorted(dictionary, key=lambda x: tuple(int(i) for i in x[key].split(".")), reverse=True)
```
umeshksingla: I believe the function name, as well as the argument name, are a little misleading as you are actually not sorting the dictionary here. 
I'd also take out the `tuple(int(i) for i in x[key].split("."))` part outside.
```
def split_version(a):
    return tuple(int(i) for i in a.split("."))

def sort_list_by_version(lst):
    return sorted(lst, key=split_version, reverse=True)
...
```
That being said, I'd actually look into writing a version comparator of our own or maybe find an implementation online somewhere in the near future.
That would totally help to avoid splitting the os_version again where you need to set the `cxx_stdlib`'s default value when the os_versions >= 10.9 or many other places perhaps.

arjunsalyan: Thank you.

arjunsalyan: > ```
> def split_version(a):
>     return tuple(int(i) for i in a.split("."))
> 
> def sort_list_by_version(lst):
>     return sorted(lst, key=split_version, reverse=True)
> ...
> ```

This won't work probably, and I am getting an error. It seems that `key` takes in a sorted list. Also, we have to sort "list" and "list of dictionaries" by some specific key. So, I believe we will need two such functions.

jmroot: > That being said, I'd actually look into writing a version comparator of our own or maybe find an implementation online somewhere in the near future.

I think distutils.version should work well enough, and is in the python standard library.

jmroot (approved):
Looks fine to me.

app/ports/utilities/sort_by_version.py
umeshksingla: While it's still not in a state what I had in mind, this certainly looks better.

app/ports/models/stats.py
```diff
@@ -38,11 +40,16 @@ def populate(cls, json_object, timestamp):
         uuid_obj, created = UUID.objects.get_or_create(uuid=json_object['id'])
         sub = Submission()
         sub.user = uuid_obj
-        sub.os_version = json_object['os'].get('osx_version')
+        os_version = json_object['os'].get('osx_version')
+        cxx_stdlib = json_object['os'].get('cxx_stdlib')
+        if cxx_stdlib is None and compare_versions(os_version, '10.9'):
```
umeshksingla: `compare_versions` as a name still doesn't tell if one is greater than the other.

arjunsalyan: Thanks, do you have some suggestion?

umeshksingla: I mean, now that we got a `LooseVersion` implemented,  you can really just do a `if ... and LooseVersion(v1) >= LooseVersion(v2)` check. 
But what was the reason behind choosing `LooseVersion` and `StrictVersion` in our case? I wasn't actually able to decide, so I don't have an answer either.

arjunsalyan: `LooseVersion` gives just some more flexibility in choosing versions as compared to `StrictVersion`. I thought that sometimes port versions don't follow a good nomenclature so using `LooseVersion` might be better.

umeshksingla: except that we are not using it for comparing port versions. But okay, I got the idea.

arjunsalyan: I have removed the function `compare_versions` until we need it somewhere else.

PR #62, opened 2019-08-01
Use transactions to load initial data

arjunsalyan:


app/ports/tests/test_dependencies.py
```diff
@@ -1,18 +1,17 @@
...
 from ports.models import Dependency, Port
 from MacPorts.config import TEST_PORTINDEX_JSON
 
 
-class TestDependencies(TestCase):
+class TestDependencies(TransactionTestCase):
+    reset_sequences = True
+
     def setUp(self):
         self.client = Client()
```
umeshksingla, line 14: I doubt you need `self.client = Client()`. It is there in the base class `SimpleTestCase`.

app/ports/tests/test_exceptions.py
```diff
@@ -1,18 +1,17 @@
...
+    reset_sequences = True
+
     def setUp(self):
         self.client = Client()
-
-    @classmethod
-    def setUpTestData(cls):
         Port.load(TEST_PORTINDEX_JSON)
 
     def test_400(self):
```
umeshksingla: Is it `test_404`?

arjunsalyan: Oh, right. thanks

app/ports/models/port.py
```diff
@@ -54,7 +54,11 @@ def load_categories_table(ports):
             Category.objects.bulk_create(batch)
             return
 
+        port_id_map = {}
+
+        @transaction.atomic
         def load_ports_and_maintainers_table(ports):
+            pid = 1
```
mojca: Why not `port_id`? `pid` may mean a lot of things.

app/ports/models/port.py
```diff
@@ -107,32 +113,31 @@ def load_ports_and_maintainers_table(ports):
...
+        @transaction.atomic
         def load_dependencies_table(ports):
-
-            def load_depends(list_of_dependencies, type_of_dependency, port):
+            def load_depends(list_of_dependencies, type_of_dependency, pid):
                 dependency = Dependency()
                 dependencies = []
                 dependency.type = type_of_dependency
-                dependency.port_name = port
+                dependency.port_name_id = pid
```
mojca: Why is this `port_name_id` if it's a number?

app/ports/models/port.py
```diff
@@ -107,32 +113,31 @@ def load_ports_and_maintainers_table(ports):
...
                 for depends in list_of_dependencies:
                     try:
-                        dependencies.append(Port.objects.get(name__iexact=depends.rsplit(':', 1)[-1]))
-                    except Port.DoesNotExist:
-                        print("Failed to append {} as a dependency to {}. Not Found.".format(depends.rsplit(':', 1)[-1],
-                                                                                             port.name))
+                        dependencies.append(port_id_map[depends.rsplit(':', 1)[-1]])
```
mojca: I would make a separate line saying something like
```python
dependency_name = depends.rsplit ...
dependencies.append(port_id_map[dependency_name])
```
to make it more clear what you are doing (one could also make three lines, but that's not needed).

app/ports/models/port.py
```diff
@@ -107,32 +113,31 @@ def load_ports_and_maintainers_table(ports):
...
+                    except KeyError:
+                        pass
                 dependency.save()
                 dependency.dependencies.add(*dependencies)
```
mojca: Just curious: why do you first save `dependency`, and only then add dependencies?

Maybe that's a requirement of django, I'm not sure?

But this makes `n` entries to the database adding `(port_id, dependency_type, dependency_id)`, right?

app/ports/models/port.py
```diff
@@ -107,32 +113,31 @@ def load_ports_and_maintainers_table(ports):
                 except KeyError:
                     pass
 
+        @transaction.atomic
         def load_dependencies_table(ports):
-
-            def load_depends(list_of_dependencies, type_of_dependency, port):
+            def load_depends(list_of_dependencies, type_of_dependency, pid):
```
mojca: Make this `port_id` as well. I would probably reverse the arguments in this function to make it somewhat more "logical": first the port id of the port that you are "modifying", then type of dependencies, then the full list of dependencies.

mojca (commented):
Some minor nitpicking comments, else it looks fine.

Is the same optimisation applicable also to maintainers, categories, ...? (It doesn't need to be done in this PR.)

app/ports/models/port.py
arjunsalyan: `port_name` is the name of the field which is a foreign key. The actual field is thus `port_name_id` which holds the id of the foreign object.

I am not sure if I got your point right?

app/ports/models/port.py
arjunsalyan: Okay, thanks.

app/ports/models/port.py
arjunsalyan: Yes, Django requires an object to be first saved before any elements are added to the M2M field.

> But this makes `n` entries to the database adding `(port_id, dependency_type, dependency_id)`, right?

The model structure is: `( port_id, dependency_type, dependecies (M2M) )`. Adding a single row hits the database twice, first it saves `(port_id, dependency_type)` and then adds the `id(s)` of dependencies to the M2M field.

arjunsalyan: I updated the naming a bit, naming the object as `dependency` was probably a bit confusing.

app/ports/models/port.py
arjunsalyan: Done, thanks.

app/ports/models/port.py
umeshksingla: Actually, I am not sure about `port_name_id` either. There could be reasons behind it because of the strict Django conventions but does it get linked to the field `port_name` using `port_name_id` above?

app/ports/models/port.py
mojca: I knew there was something fishy about the dependency design. I didn't fully understand the code in python, so I checked the database. At the moment we have two tables:

* `ports_dependency`:
  * `id` = primary key
  * `type`
  * `port_name_id` = foreign key to `port_id`
* `ports_dependency_dependencies`
  * `id` = primary key
  * `dependency_id` = foreign key to `ports_dependency_id`
  * `port_id` = foreign key to `port_id`

and in addition we have two independent (but same) indices on the same field on the same table.

I would like to reduce this overly complicated (and as a consequence slow) design to a single table:
* `ports_dependencies`:
  * `port_id` = foreign key to `port_id`
  * `type`
  * `dependency_id` = foreign key to `port_id`
with:
* index on `port_id`
* index on `type`
* index on `dependency_id`
* unique constraint on the tripple (`port_id`, `type`, `dependency_id`)
Maybe Django will insist in another id for that table. If so, let it be, but it won't be used anywhere.

I'm pretty sure that this should speed up the initial data loading and that there should be no need to store the object in two phases. One should be able to simply fill the table. Fast.

arjunsalyan: The [current model](https://github.com/arjunsalyan/macports-gsoc-2019-webapp/blob/f26dd8613936e98ccbcba34c38622550f676ca04/app/ports/models/port.py#L295) is defined as:
- `port_name` - foreign key to `port_id`.
- `type`
- `dependencies` - many to many field with `port_id`

Rest everything is done by Django. We have two tables because one of the tables holds relations for the many-to-many mapping, and is handled completely by Django.

> I would like to reduce this overly complicated (and as a consequence slow) design to a single table:
> 
> * `ports_dependencies`:
>   
>   * `port_id` = foreign key to `port_id`
>   * `type`
>   * `dependency_id` = foreign key to `port_id`

Thank you. In one way or other, we have the same design. Just the last field: `dependencies` is a M2M field. The proposed table might make the process of updating a bit more complex, because all the previous entries will have to be deleted- while currently we simply clear the relations. Also, it seems that it will take more space on the database, with around 50,000 rows filled with the string: `type`. Sorry, if I am totally wrong about this, but these were the points I had in my mind while making this model.

>     with:
> * index on `port_id`
> * index on `type`
> * index on `dependency_id`
> * unique constraint on the tripple (`port_id`, `type`, `dependency_id`)
>   Maybe Django will insist in another id for that table. If so, let it be, but it won't be used anywhere.

Thank you, in any case the indexing for this Model can be improved.

> I'm pretty sure that this should speed up the initial data loading and that there should be no need to store the object in two phases. One should be able to simply fill the table. Fast.

Okay, I will test with the proposed table design and will report the results.

mojca: Can you let me know the precise sql structure if you specify the many-to-many relationship in [this way](https://docs.djangoproject.com/en/2.2/ref/models/fields/#django.db.models.ManyToManyField.symmetrical)?

```python
class Port(models.Model):
    # ...
    dependencies = models.ManyToManyField("self", symmetrical=False)
```

The type of dependency might be specified using [though_fields](https://docs.djangoproject.com/en/2.2/ref/models/fields/#django.db.models.ManyToManyField.through_fields) if necessary.

If you want to reduce the characters stored for the type of dependency (`run`, `lib`, ...), we could actually some python enum and never actually retrieve those values from the database. Say: lib = 0, build = 1, run = 2, fetch = 3, ... and only operate with those numbers on the database level, while the string itself would only ever be printed when generated the website, from the static python array.

(Maybe `replaced_by` could have a similar one-to-one treatment, even though I'm not entirely sure about it due to potential complications with having to have the replaced_by port available before the obsolete one.)

arjunsalyan: > Can you let me know the precise sql structure if you specify the many-to-many relationship in [this way](https://docs.djangoproject.com/en/2.2/ref/models/fields/#django.db.models.ManyToManyField.symmetrical)?

Not sure of the sql structure, I have never used this. I can read more in the evening. But do we want such a relation: if A is a dependency of B, then B will be a dependency of A?

From understanding point of view, the current implementation seems to be the most straightforward, it is how such relations are generally handled in Django.

Thank you for the references and suggestions, I will look into a better way to store dependencies.

arjunsalyan: > I would like to reduce this overly complicated (and as a consequence slow) design to a single table:
> 
> * `ports_dependencies`:
>   
>   * `port_id` = foreign key to `port_id`
>   * `type`
>   * `dependency_id` = foreign key to `port_id`

I tested out with this design, and the time taken was almost the same. I expected it to increase because of the increase in the number of hits to the database but it remained the same, which might be due to the use of transactions and `port_id_map`. Shall we proceed with this design then?
How shall we take care of `ports.macports.org`, reload the data or use migration to populate the new table from the old one? Some significant changes will also have to be done in the updating of port info, views and tests.

mojca: Let's merge this one as it is for now and proceed with API etc., then look into this again at the end.

Maybe the easiest migration would be to drop the relevant tables and either create tables again from scratch (if empty) or re-read portindex.json. This is generally a bad practice, but our statistics & build tables are luckily independent, so it shouldn't be an issue.

arjunsalyan: Thank you.

PR #64, opened 2019-08-03
All commits for general feedback

mojca:
This is a "fake pull requests" to simplify general feedback about the code.


app/ports/models/stats.py
```diff
@@ -38,11 +40,16 @@ def populate(cls, json_object, timestamp):
         uuid_obj, created = UUID.objects.get_or_create(uuid=json_object['id'])
         sub = Submission()
         sub.user = uuid_obj
-        sub.os_version = json_object['os'].get('osx_version')
+        os_version = json_object['os'].get('osx_version')
+        cxx_stdlib = json_object['os'].get('cxx_stdlib')
+        if cxx_stdlib is None and LooseVersion(os_version) >= LooseVersion('10.9'):
```
mojca, line 47: What if `os_version` is `None` / what if this is not macOS?

We probably want a test to make sure it behaves correctly.

app/ports/models/port.py
```diff
@@ -0,0 +1,335 @@
...
+    variant = models.CharField(max_length=100, default='')
+
+
+class Maintainer(models.Model):
+    name = models.CharField(max_length=50, default='')
+    domain = models.CharField(max_length=50, default='')
+    github = models.CharField(max_length=50, default='')
+    ports = models.ManyToManyField(Port, related_name='maintainers')
+
+    objects = PortManager()
```
umeshksingla, line 327: Is it okay to use the same manager for both the `Port` model as well as `Maintainer`? Will `name` be able to uniquely identify a maintainer? I'd go with separate ones in case we end up modifying manager class to suit needs for the `Port` model and overlooking `Maintainer` model fields.
I may have overlooked the use case around using a manager at all in our case.

app/ports/models/buildhistory.py
```diff
@@ -0,0 +1,127 @@
...
+
+from MacPorts.config import BUILDERS_JSON_URL, BUILDBOT_URL_PREFIX, BUILDS_FETCHED_COUNT
+
+
+class Builder(models.Model):
+    name = models.CharField(max_length=100, db_index=True)
+
+
+class BuildHistory(models.Model):
+    builder_name = models.ForeignKey(Builder, on_delete=models.CASCADE)
```
mojca, line 17: If we are storing the `id`, we should call this field `builder`, rather than `builder_name` (and then end up with `builder_name_id`) as the latter is confusing.

mojca: I'm also confused about why we end up with two indices:
```
CREATE INDEX ports_build_builder_289e95_idx
    ON public.ports_buildhistory USING btree
    (builder_name_id)
    TABLESPACE pg_default;

CREATE INDEX ports_buildhistory_builder_name_id_9ac9bd18
    ON public.ports_buildhistory USING btree
    (builder_name_id)
    TABLESPACE pg_default;
```

app/ports/models/port.py
```diff
@@ -0,0 +1,335 @@
...
+        return self.get(name=name)
+
+
+class Port(models.Model):
+    portdir = models.CharField(max_length=100)
+    description = models.TextField(default='')
+    homepage = models.URLField(default='')
+    epoch = models.BigIntegerField(default=0)
+    platforms = models.TextField(null=True)
+    categories = models.ManyToManyField(Category, related_name='category', db_index=True)
```
mojca, line 30: This now creates a column `category_id` which suggests a numeric value, but actually holds a text value with full category name.

The full table is
* `id` = primary key (not really needed anywhere, I assume?)
* `port_id` = foreign key to `port_id`; numeric
* `category_id` = foreign key to `category_name`; text

The last two fields look somewhat inconsistently used.

app/ports/models/stats.py
```diff
@@ -0,0 +1,90 @@
+from distutils.version import LooseVersion
+
+from django.db import models
+from django.contrib.postgres.fields import JSONField
+
+
+class UUID(models.Model):
+    uuid = models.CharField(max_length=36, db_index=True)
```
mojca, line 8: `uuid` needs to be unique.

mojca:
Closing it now as it's no longer relevant.

PR #67, opened 2019-08-05
models.stats: Add clt_version and variants fields

arjunsalyan:


mojca (approved):

arjunsalyan:
This PR is probably complete now.

app/ports/migrations/0006_add_clt_version_and_variants.py
```diff
@@ -0,0 +1,31 @@
...
+
+    dependencies = [
+        ('ports', '0005_remove_models_user_osdist'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='portinstallation',
+            name='variants',
+            field=models.CharField(default='', max_length=200),
```
mojca, line 16: I cannot really decide how much we need here, and would be tempted to either use perhaps 256 or just text with no limits. But if we use text, maybe the performance could decrease. Maybe 200 is sufficient?

app/ports/models/stats.py
```diff
@@ -17,6 +17,7 @@ class Submission(models.Model):
     platform = models.CharField(max_length=20, default='')
     macports_version = models.CharField(max_length=10)
     cxx_stdlib = models.CharField(max_length=20, default='libc++')
```
mojca: Didn't we say that we wouldn't put any defaults here?

app/ports/migrations/0007_alter_field_clt_version.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('ports', '0006_add_clt_version_and_variants'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='submission',
+            name='clt_version',
+            field=models.CharField(default='', max_length=100),
```
mojca, line 16: Do we need two migrations in the same PR? (Maybe we do because of the deployed version.)

mojca (approved):

app/ports/migrations/0007_alter_field_clt_version.py
arjunsalyan: Yes, the deployed demo version needs that we put these migration files.

app/ports/models/stats.py
arjunsalyan: Oh, thank you very much. It is from a previous commit and got overlooked.

app/ports/migrations/0007_alter_field_clt_version.py
arjunsalyan: We now have three. But there is a way to avoid this, we can replace them with a single file and then run a fake migration.

mojca: I leave that up to you. I'm fine with any solution, so feel free to merge it yourself in the proper way.

mojca (commented):
As I already said last time, I'm ok with the changes, so feel free to press the merge button.

There's one missing piece that could be changed. Running a query that fetches all old submissions which lack `clt_version` or `variants` field, but contain `raw_json` field, and fill in the missing information.

That's somewhat optional and can be done in a separate PR to avoid postponing this one. But we might need something similar later as well, so it could be useful to set an example now. On the deployed version you can no longer manually change the database.

PR #68, opened 2019-08-05
Fix exception and searching for maintainers page

### LINKED ISSUE #63: Broken maintainer's page when invalid maintainer handle
We need to rethink a little around the maintainer's page. I was of the opinion that when clicked on a maintainer's name in the port-summary page, it'll simply go to another page but instead, it tries to load the content over it. (Browser's backward and forward buttons don't work with the maintainer's page, to verify). We really don't need any in-place modification of content like this.

Also, passing a wrong maintainer's handle breaks the app on accessing `all_ports` variable, e.g. http://\<ec2\>/maintainer/github/`umesh`.  The correct URL was http://\<ec2\>/maintainer/github/`umeshksingla`. I don't think the query gets fired.

arjunsalyan:
Closes: https://github.com/macports/macports-webapp/issues/63

### DIFF
--- a/app/ports/views.py
+++ b/app/ports/views.py
@@ -390,6 +390,8 @@ def get_ports_of_maintainers(maintainers, request):
 
 def maintainer_detail_github(request, github_handle):
     maintainers = Maintainer.objects.filter(github=github_handle)
+    if maintainers.count() == 0:
+        return render(request, '404.html')
     ports, all_ports_num = get_ports_of_maintainers(maintainers, request)
 
     return render(request, 'ports/maintainerdetail.html', {
@@ -403,6 +405,8 @@ def maintainer_detail_github(request, github_handle):
 
 def maintainer_detail_email(request, name, domain):
     maintainers = Maintainer.objects.filter(name=name, domain=domain)
+    if maintainers.count() == 0:
+        return render(request, '404.html')
     ports, all_ports_num = get_ports_of_maintainers(maintainers, request)
 
     return render(request, 'ports/maintainerdetail.html', {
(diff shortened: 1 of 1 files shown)


app/ports/views.py
```diff
@@ -468,9 +472,9 @@ def search_ports_in_category(request):
 # Respond to ajax calls for searching within a maintainer
 def search_ports_in_maintainer(request):
     query = request.GET.get('name', '')
-    name = request.GET.get('maintainers__name')
+    name = request.GET.get('maintainers__name', '')
     github = request.GET.get('maintainers__github')
-    if name is None:
+    if name is None or name == '':
```
umeshksingla, line 477: Will `name` ever be `None`?

app/ports/views.py
```diff
@@ -403,6 +405,8 @@ def maintainer_detail_github(request, github_handle):
 
 def maintainer_detail_email(request, name, domain):
     maintainers = Maintainer.objects.filter(name=name, domain=domain)
+    if not maintainers.count() > 0:
```
umeshksingla: I think we can write it in a more straightforward way which to me is more readable: `if maintainers.count() == 0`. What do you think?

arjunsalyan: Seems fine to me.

app/ports/views.py
arjunsalyan: Probably never, but I thought it won't hurt to have it. Shall I remove this?

umeshksingla: It should be fine. Though I hope we don't incur a case where both `name` and `github` are `''`.

umeshksingla (approved):

app/ports/views.py
jmroot: That would be a mistake in the portfile, but you should make sure you don't break if it does happen.

PR #72, opened 2019-08-06
Travis: Test building of docker image and starting container

arjunsalyan:


mojca:
There were connectivity issues during the build. We should try to re-trigger the builds in one way or another.

env.sample
```diff
@@ -1,6 +1,6 @@
-DB_NAME=
-DB_USER=
+DB_NAME=mp
+DB_USER=postgres
 DB_PASSWORD=
-DB_HOST=
+DB_HOST=localhost
 DB_PORT=
-SECRET_KEY=
+SECRET_KEY=aRandomKey
```
mojca: Given that you already specified some values in the Travis configuration script, could the same values be reused somehow? (It's not absolutely necessary to change it, just potentially nice.)

Other than that, it looks fine to me.

arjunsalyan:
I restarted the build, it has passed.

env.sample
arjunsalyan: Thank you, I have done the changes.

mojca (approved):
Even if there is some better way to check for running image, this commit is helpful.

PR #73, opened 2019-08-07
Do not generate migration files inside the container

arjunsalyan:
I just got a realisation that if somehow a migration file is generated inside the container and applied to the database (which is not there in the repository), it may raise conflicts that will require the server administrator's help to fix.

So, the better way would be to supply the migration files when changes are done or whenever required and stop generating any such migrations inside the container. I am not totally sure about this, or do we have any other way to deal with such production environments?

### DIFF
--- a/config/setup_app.sh
+++ b/config/setup_app.sh
@@ -2,8 +2,6 @@
 
 cd /code/app
 
-python3 manage.py makemigrations --noinput
-
 python3 manage.py migrate --noinput
 
 python3 manage.py collectstatic --noinput


mojca (approved):
Good point. Looks ok.

umeshksingla:
So, as far as I understand here, there can be conflicts with migration files sometimes but never during actual db migrations which need a human intervention?

Any example case for migration file conflicts? (unless we start force pushing migration files as well).

mojca:
I can imagine that accidentally a migration file under some random name might be created inside the container. But when we then add further migrations later, they probably reference the name of the last applied migration. And that one could be different in the docker than in the repository.

I assume this would only happen if someone would forget to commit the migration alongside the database changes. But when I think of it, this would probably also generate issues? What happens (before and after this proposed change) if you modify the database and forget to commit migrations alongside the model change?

arjunsalyan:
Yes, in case someone forgets to commit the migrations alongside changes to models, then the docker container would generate and apply that migration.

Now, when the container is restarted, the file will be cleared. Later if someone commits the migrations that are not in the repository, the container would try to apply the migration which has already been applied, and will fail.

When the changes done are not applied to the database, only the modified functionality would break, and might be a good signal about what is missed.

PR #77, opened 2019-08-08
Add table of variants and clt_version

arjunsalyan:
- [x] Add table of variants
- [x] Add table/chart of clt_verison

### DIFF
--- a/app/ports/templates/ports/port-detail/installation_stats.html
+++ b/app/ports/templates/ports/port-detail/installation_stats.html
@@ -69,6 +69,11 @@
             </table>
         </div>
     </div>
+    <div class="row">
+        <div class="col-lg-6">
+            <div class="chart" id="clt-version-chart"></div>
+        </div>
+    </div>
 
     <br><br>
     <h4 class="border-bottom">Port installations by month <span style="font-size:12px"><button type="button" class="bg-light border-0 p-1" data-toggle="modal" data-target="#monthlyInstallationsHelp">Learn More</button></span> </h4>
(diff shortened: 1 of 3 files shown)


app/ports/templates/ports/port-detail/installation_stats.html
```diff
@@ -256,10 +283,40 @@ <h4 class="modal-title">How are monthly installations calculated?</h4>
...
+
+            var data = google.visualization.arrayToDataTable([
+                ['OS Version', {% for version in clt_parent %}'{% if version.grouper == '' %}unknown{% else %}{{ version.grouper }}{% endif %}', {% endfor %} { role: 'annotation' } ],
+                {% for os in os_clt_parent %}
+                    ['OS: {{ os.grouper }}', {% for version in clt_parent %}{% for i in os.list %}{% if i.submission__clt_version == version.grouper %}{{ i.num }}{% endif %}{% endfor %}, {% endfor %} ''],
+                {% endfor %}
+        ]);
+
+        var options = {
+            'title': 'CLT Versions',
```
mojca: Let's spell that name out.

I would use singular for version all over the place, but that's unrelated

arjunsalyan: Okay, thanks.

mojca:
Here's how the wrong sorting order of CLT (for the ld64 port) looks like:

image

mojca:
I see that `xcode` uses `sortversion`:
```
{% regroup os_version_and_xcode_version|sortversion:"xcode_version" by xcode_version as xcode_parent %}
```
while clt uses `dictsort`:
```
{% regroup os_version_and_clt_version|dictsort:"clt_version" by clt_version as clt_parent %}
```

PR #80, opened 2019-08-13
Add tests for build history and remove initiating of Client()

arjunsalyan:


mojca:
I don't have any further suggestion for this (honestly I'm not sure about all the details), but the conflict needs to be resolved. If @umeshksingla doesn't have anything to add, feel free to merge.

app/ports/tests/test_buildhistory.py
```diff
@@ -0,0 +1,45 @@
...
+class TestDependencies(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        Port.load(TEST_PORTINDEX_JSON)
+
+    def test_loading_builders(self):
+        BuildHistory.populate_builders()
+
+        loaded = False
+        if Builder.objects.count() >= 11:
```
umeshksingla, line 15: I think you can at best check for a non-zero number of objects populated in `BuildHistory` here. We don't know if >=11 is going to be fixed, do we (@mojca)?
I think we should rather test this against a fixed JSON file with the format we expect from the upstream.

app/ports/tests/test_buildhistory.py
```diff
@@ -0,0 +1,45 @@
...
+            loaded = True
+        self.assertEquals(loaded, True)
+
+    def test_fetch_and_filter_builhistory(self):
```
umeshksingla: typo in `...builhistory`

app/ports/tests/test_buildhistory.py
```diff
@@ -0,0 +1,45 @@
...
+from MacPorts.config import TEST_PORTINDEX_JSON, BUILDS_FETCHED_COUNT
+
+
+class TestDependencies(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        Port.load(TEST_PORTINDEX_JSON)
+
+    def test_loading_builders(self):
+        BuildHistory.populate_builders()
```
umeshksingla, line 12: This is of the least importance. But do you also think that it just seems like a wrong abstraction of having `populate_builders` method in `BuildHistory` model?

app/ports/tests/test_buildhistory.py
umeshksingla: Or is the test intended to test the buildbot API call?

app/ports/tests/test_buildhistory.py
```diff
@@ -0,0 +1,45 @@
+import os
+
+from django.test import TestCase
+from django.urls import reverse
+
+from ports.models import Dependency, Port, BuildHistory, Builder
+from MacPorts.config import TEST_PORTINDEX_JSON, BUILDS_FETCHED_COUNT
+
+
+class TestDependencies(TestCase):
```
umeshksingla: Does this not need be a `TransactionTestCase` with `reset_sequences` enabled?

arjunsalyan: This does not need, but for consistency, I would use `TransactionTestCase` here also.

app/ports/tests/test_buildhistory.py
arjunsalyan: Thank you.

app/ports/tests/test_buildhistory.py
arjunsalyan: I just thought that we currently have 11 active builders and this number will only increase with time so I used this. But I am not sure.

app/ports/tests/test_buildhistory.py
arjunsalyan: Hmm. It should have been `Builder.populate_builders` in the `Builder` model.

arjunsalyan: Shall I change this in this PR only?

umeshksingla:
No, let’s keep it for later. It’s a model change. I just wanted to know
your opinion as well.

app/ports/tests/test_buildhistory.py
umeshksingla: No, I didn't mean for consistency across the project but really the tests itself.

app/ports/tests/test_buildhistory.py
umeshksingla: No, let’s keep it for later. It’s a model change. I just wanted to know
your opinion as well.

app/ports/tests/test_buildhistory.py
umeshksingla: Let's ask @mojca.

arjunsalyan:
Some improvements were done to how we handle build history and related tests were added. This PR can be closed now.

PR #81, opened 2019-08-13
Project Documentation

arjunsalyan:


docs/3-MODELS.md
```diff
@@ -0,0 +1,124 @@
+## 2. MODELS (Storing the collected data)
+
+The data collected from various sources, as discussed in [2-SOURCES](/docs/2-SOURCES.md), is stored in a PostgreSQL
+database. The tables are designed using Django models.
+
+All of the project's models are defined inside the app: `ports`. The models are split into three different files: `port`,
+`buildhistory` and `stats`. Each file contains the models related to the category defined by file name.
```
mojca: These could also be "internal links" (`ports`, `port`, `buildhistory`, `stats`), so that user reading the markdown on the webpage could just click and go straight to the file in question.

docs/3-MODELS.md
```diff
@@ -0,0 +1,124 @@
...
+- `portdir`
+- `description`
+- `homepage`
+- `epoch`
+- `platforms`
+- `categories` - A many-to-many relation with `Category` model
+- `long_description`
+- `version`
+- `revision`
+- `closedmaintainer`
```
mojca: It's probably makes sense to add a sentence describing what this is since our nomenclature is completely non-standard.

docs/3-MODELS.md
```diff
@@ -0,0 +1,124 @@
...
+The data collected from various sources, as discussed in [2-SOURCES](/docs/2-SOURCES.md), is stored in a PostgreSQL
+database. The tables are designed using Django models.
+
+All of the project's models are defined inside the app: `ports`. The models are split into three different files: `port`,
+`buildhistory` and `stats`. Each file contains the models related to the category defined by file name.
+
+#### 1. `Port` Model
+
+It stored the basic information related to a port, with the following fields:
+- `portdir`
```
mojca: Maybe a short example as it's not necessarily clear to everyone what this is. I would also add:

> Subports share the same `portdir`.

docs/3-MODELS.md
```diff
@@ -0,0 +1,124 @@
+## 2. MODELS (Storing the collected data)
+
+The data collected from various sources, as discussed in [2-SOURCES](/docs/2-SOURCES.md), is stored in a PostgreSQL
+database. The tables are designed using Django models.
+
+All of the project's models are defined inside the app: `ports`. The models are split into three different files: `port`,
+`buildhistory` and `stats`. Each file contains the models related to the category defined by file name.
+
+#### 1. `Port` Model
```
mojca: I would either use just "Port" (probably without backticks) or "The Port Model". (Doesn't English always require as article before nouns?)
I don't see the need to use backticks in the title this case.
A matter of personal preference though.

docs/3-MODELS.md
```diff
@@ -0,0 +1,124 @@
...
+- `description`
+- `homepage`
+- `epoch`
+- `platforms`
+- `categories` - A many-to-many relation with `Category` model
+- `long_description`
+- `version`
+- `revision`
+- `closedmaintainer`
+- `name`
```
mojca: I feel that `name` is the most important field. I would put it on top.

docs/3-MODELS.md
```diff
@@ -0,0 +1,124 @@
...
+
+#### 1. `Port` Model
+
+It stored the basic information related to a port, with the following fields:
+- `portdir`
+- `description`
+- `homepage`
+- `epoch`
+- `platforms`
+- `categories` - A many-to-many relation with `Category` model
```
mojca: What I find confusing this in this description overall (but it might be related to our basic database design/description) is that all of the following fields:
* maintainer
* categories
* variants
are all many-to-many relationships, yet each one of them is presented differently.

docs/3-MODELS.md
```diff
@@ -0,0 +1,124 @@
...
+- `long_description`
+- `version`
+- `revision`
+- `closedmaintainer`
+- `name`
+- `license`
+- `replaced_by` - Default values is `Null` for active ports. If the value is not null, the port is considered as obsolete.
```
mojca:
```suggestion
- `replaced_by` - `NULL` by default. If the port is obsolete and replaced by another one, the name of the new port is stored here.
```

docs/3-MODELS.md
```diff
@@ -0,0 +1,124 @@
...
+- `active` - Value is `False` for deleted ports, otherwise `True`*.
```
mojca:
```suggestion
- `active` - `True` by default. `False` when the port no longer exists (in `PortIndex`).
```

docs/3-MODELS.md
```diff
@@ -0,0 +1,124 @@
...
+
+***A Port is marked as deleted when the port is available in the app's database, but does not exist in PortIndex. If a 
+port with the same name is added later in time, then previous object is replaced with new object and the newer ports gets
+marked as `active=True` automatically.*
```
mojca: This way of using kind-of-footnotes is a bit strange (fist you have one star, then two for the actual description; and the star could interfere with markup, ...).
I would have pointed you [here](https://stackoverflow.com/questions/25579868/how-to-add-footnotes-to-github-flavoured-markdown), but if you look at the other document, maybe you could simply add notes at the end of each table (when there's a reason to do so), avoiding footnotes altogether.

docs/3-MODELS.md
mojca: Could you provide an example (URL pointing to ports.macports.org) of both (one `replaced_by` and one `active=False`) ports? For the `replaced_by` pick some python or perl subport which is simply replaced by a newer subport.

docs/3-MODELS.md
```diff
@@ -0,0 +1,124 @@
...
+Stores email and github-handle of the maintainers, and maps them with objects from the `Port` model.
+
+- `name`
+- `domain`
+- `github`
+- `ports`- many-to-many relation with the `Port` Model.
+
+*The `name`, `domain`, `github` triplet is unique. But due to inconsistencies in the `Portfiles`, there might be cases
+when we have two entries with same github-handle but different emails. In such cases, a warning is displayed on the
+maintainer's page.*
```
mojca: I'm not sure if I ever saw such a warning. But what would be more useful would be a page displaying all such inconsistencies, so that we can easily fix them, else it's pretty impossible to find one. Not super high priority, but something to note down.

docs/3-MODELS.md
```diff
@@ -0,0 +1,124 @@
...
+- `variant`
+
+
+#### 5. `Dependency` Model
+
+Contains relation for ports and their dependencies:
+
+- `port_name` - foreign key to `Port` model
+- `dependencies` - many-to-many relation with `Port` model
+- `type`
```
mojca: Write a list of valid values for `type`.

docs/3-MODELS.md
```diff
@@ -0,0 +1,124 @@
...
+
+
+#### 6. `LastPortIndexUpdate` Model
+
+Stores only one object. The object tells about the git commit till which the port information is up-to-date and the time
```
mojca: Maybe:
```suggestion
Stores only one object containing the shasum of the git commit from which we extracted the latest information about ports.
```

docs/3-MODELS.md
```diff
@@ -0,0 +1,124 @@
...
+
+- `git_commit_hash`
+- `updated_at`
+
+---
+
+#### 7. `Builder` Model
+
+Stores names of the active builders from Buildbot. The build history is fetched only for these builders. This is the only
+place where information about builders is stored.
```
mojca, line 110: Do we explain anywhere how this table is filled?

docs/3-MODELS.md
```diff
@@ -0,0 +1,124 @@
...
+
+- `builder_name` - foreign key to `Builder` Model
+- `build_id`
+- `status`
+- `port_name`
+- `time_start`
+- `time_elapsed`
+- `watcher_id`
+
+---
```
mojca, line 126: This could be a new section (see the other old document).

docs/3-MODELS.md
```diff
@@ -0,0 +1,124 @@
...
+
+#### 9. `UUID` Model
+
+Stores UUID of the users who have made submissions. There are no duplicates in the table, no matter how many submissions
+one user makes there is only one object representing the user.
```
mojca: You should still print the table with columns.

docs/3-MODELS.md
```diff
@@ -0,0 +1,124 @@
...
+- `os_version`
+- `xcode_version`
+- `os_arch`
+- `build_arch`
+- `platform`
+- `macports_version`
+- `cxx_stdlib`
+- `clt_version`
+- `raw_json` - stores the entire JSON object received in the submission
+- `timestamp`
```
mojca: Add a note that this is the timestamp on the server when the submission was received. (Everything else comes from submission.)

mojca (commented):
Can you please also look at https://github.com/macports/macports-gsoc2018-webapp/blob/master/docs/Database_Design.md and steal some ideas?

What I like about that other document is:
* table of contents
* data model / columns presented as a table rather than items (I find the table easier to interpret)
* I have a feeling that we are slightly misusing backticks :); that page has less of them
* more structured description about what's in the data (separately name, type, description, ...)

docs/3-MODELS.md
arjunsalyan: I was thinking of doing this, thanks.

docs/3-MODELS.md
arjunsalyan: Okay, thanks.

docs/3-MODELS.md
arjunsalyan: Maintainer, Dependencies and Variants have separate tables, with a foreign key relating to the `Port` model.

I guess I do not completely understand what you are trying to say here.

docs/3-MODELS.md
arjunsalyan: Here is an example: https://ports.macports.org/maintainer/github/yan12125/

PR #82, opened 2019-08-14
git_update.py: Do not print to stderr on success

neverpanic:
We use a script that sends emails for cron job runs if they exited with a non-zero error code or printed something on stderr. Unfortunately, git's progress information is printed to stderr, which means we currently get an email for every update run.

Silence this by passing `--quiet` to git in a couple of places.

### DIFF
--- a/app/parsing_scripts/git_update.py
+++ b/app/parsing_scripts/git_update.py
@@ -22,7 +22,7 @@ def clone_repo():
     if os.path.isdir(MACPORTS_PORTS_DIR):
         return
     else:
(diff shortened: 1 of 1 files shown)


arjunsalyan:
Thank you. We are also printing at several places using `print()` in the update process. I believe we only receive emails for `stderr`, then these print statements should not be a problem?

neverpanic:
I get the entire output mailed but only if there was output on stderr or the command failed. Consequently, it is OK (and even useful) to have output on stdout.

PR #83, opened 2019-08-14
Exclude 'mpstats' and 'mpstats-gsoc' from top ports

arjunsalyan:


app/ports/views.py
```diff
@@ -23,7 +23,7 @@ def index(request):
     categories = Category.objects.all().order_by('name')
 
     submissions_unique = Submission.objects.filter(timestamp__gte=datetime.datetime.now(tz=datetime.timezone.utc) - datetime.timedelta(days=30)).order_by('user', '-timestamp').distinct('user')
-    top_ports = PortInstallation.objects.filter(submission_id__in=Subquery(submissions_unique.values('id')), requested=True).exclude(port__exact='mpstats-gsoc').values('port').annotate(num=Count('port')).order_by('-num')[:10]
+    top_ports = PortInstallation.objects.filter(submission_id__in=Subquery(submissions_unique.values('id')), requested=True).exclude(port__in=['mpstats', 'mpstats-gsoc']).values('port').annotate(num=Count('port')).order_by('-num')[:10]
```
mojca: It is ok as it is.

My slight personal personal preference would be to match on (the django equivalent of):

```sql
WHERE port LIKE 'mpstats%'
```

which would match any port starting with `mpstats`.

Untested:
```suggestion
    top_ports = PortInstallation.objects.filter(submission_id__in=Subquery(submissions_unique.values('id')), requested=True).exclude(port__startswith='mpstats').values('port').annotate(num=Count('port')).order_by('-num')[:10]
```

app/ports/views.py
```diff
@@ -349,7 +349,7 @@ def stats_port_installations_filter(request):
         .filter(submission_id__in=Subquery(submissions_unique.values('id')))\
         .values('port').annotate(total_count=Count('port'))\
         .annotate(req_count=Count(Case(When(requested=True, then=1), output_field=IntegerField())))\
-        .exclude(port__iexact='mpstats-gsoc')\
+        .exclude(port__in=['mpstats', 'mpstats-gsoc'])\
```
mojca: Same as above.

mojca (commented):
It looks ok, just a minor optional suggestion. Either change it or not, then feel free to simply merge yourself.

mojca (approved):

app/ports/views.py
arjunsalyan: I had this in the mind initially, but was not sure, now I can do this definitely. Thank you.

PR #84, opened 2019-08-14
Handle exceptions while connecting to external sources

arjunsalyan:
- [x] Handle exception when connecting to buildbot fails while fetching build history
- [x] Handle exception when connecting to rsync server fails

### DIFF
--- a/app/ports/models/buildhistory.py
+++ b/app/ports/models/buildhistory.py
@@ -45,9 +45,12 @@ class BuildHistory(models.Model):
 
         def get_data_from_url(url):
             gcontext = ssl.SSLContext()
-            with urllib.request.urlopen(url, context=gcontext) as u:
-                data = json.loads(u.read().decode())
-            return data
+            try:
+                with urllib.request.urlopen(url, context=gcontext) as u:
+                    data = json.loads(u.read().decode())
+                return data
+            except urllib.error.URLError:
+                return {}
 
         def get_build_properties(array):
             properties = {}
@@ -94,6 +97,8 @@ class BuildHistory(models.Model):
         for buildername in builders:
             # fetch the last build first in order to figure out its number
             last_build_data = get_data_from_url(get_url_json(buildername, -1))
+            if not last_build_data:
+                continue
             last_build_number = last_build_data['number']
             build_number_loaded = BuildHistory.objects.filter(builder_name__name=buildername).order_by('-build_id')
             if build_number_loaded:
@@ -103,6 +108,8 @@ class BuildHistory(models.Model):
 
             for build_number in range(build_in_database, last_build_number):
                 build_data = get_data_from_url(get_url_json(buildername, build_number))
+                if not build_data:
+                    break
                 build_data_summary = return_summary(buildername, build_number, build_data)
                 load_database(build_data_summary)
 
--- a/app/ports/management/commands/autoload-initial-data.py
+++ b/app/ports/management/commands/autoload-initial-data.py
@@ -14,6 +14,9 @@ class Command(BaseCommand):
         # Fetch the latest version of PortIndex.json and open the file
         data = Port.PortIndexUpdateHandler().sync_and_open_file()
 
+        if not data:
+            return
+
         # Start populating
         Port.load(data['ports'])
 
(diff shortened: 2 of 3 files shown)


neverpanic (commented):
LG2M, but I'm no expert on this. Will another attempt be made to obtain the results of the builds (e.g. if the server temporarily answers with HTTP 500)?

mojca:
There is generally a cron job that runs on regular basis (I'm not sure how often, but I could check). The latest successfully imported commit (to update the ports) is saved to the database and whenever the next update happens, only paths that have been modified between the latest recorded commit and the one stored in `portindex.json` are considered for update. So if one attempt fails, it should update again when the cron job runs next time. I'm not sure if we are properly protecting everything with transactions yet, but if fetching the data fails, the ports should still all be properly up to date next time when the job is started.

The build history also starts fetching the history from the last database entry on (I hope). I never double-checked how cases with unfinished builds fetched are treated, but in principle some failed command should not be a problem. What may need double-checking is that we fetch the builds in the correct order. That is: if we see that 10 builds are missing in our database, for example builds 310 to 319, we should make sure that we don't store 310, skip 311, store 312, skip 313 and newer, as in that case we might never retry to fetch the build 311. Or that we don't fetch 319 down to 315, after which 314 fails and we never retry 310-314. @arjunsalyan please check that (I'll check it as well).

It might be nice to implement some "internal database health checks" and make sure that there are no holes in builds for any of the builders.

app/ports/models/buildhistory.py
```diff
@@ -103,6 +108,8 @@ def load_database(data):
 
             for build_number in range(build_in_database, last_build_number):
                 build_data = get_data_from_url(get_url_json(buildername, build_number))
+                if not build_data:
+                    continue
```
mojca: Reading this code I suspect that if fetching just one build out of ten would fail, we'll happily continue to the next number, but we will never recover later (never attempt to fetch that one build again) and we'll always keep the hole in the data.

Please correct me if I'm wrong.

arjunsalyan: Oh, that's right, thank you. I believe a `break` would be a better choice, because the app is very well capable of recovering from the last state.

arjunsalyan:
> There is generally a cron job that runs on regular basis (I'm not sure how often, but I could check). The latest successfully imported commit (to update the ports) is saved to the database and whenever the next update happens, only paths that have been modified between the latest recorded commit and the one stored in `portindex.json` are considered for update. So if one attempt fails, it should update again when the cron job runs next time. I'm not sure if we are properly protecting everything with transactions yet, but if fetching the data fails, the ports should still all be properly up to date next time when the job is started.

Currently, the commit is saved to the database only when the entire update process ran successfully.

> The build history also starts fetching the history from the last database entry on (I hope). I never double-checked how cases with unfinished builds fetched are treated, but in principle some failed command should not be a problem. What may need double-checking is that we fetch the builds in the correct order. That is: if we see that 10 builds are missing in our database, for example builds 310 to 319, we should make sure that we don't store 310, skip 311, store 312, skip 313 and newer, as in that case we might never retry to fetch the build 311. Or that we don't fetch 319 down to 315, after which 314 fails and we never retry 310-314. @arjunsalyan please check that (I'll check it as well).

If the highest build id in the database is 310, next time it would start with 311. The highest build id is the only way by which it recognises where to start. But the recent change will definitely leave holes and needs to be improved.

app/ports/models/buildhistory.py
arjunsalyan: I have made the changes, and now the flow is: 

1. We loop over all the builders.
2. For each builder, we try to fetch the most recent build, if the process fails, we continue to next builder.
3. If it succeeds, we obtain the highest build id from the database. And loop from 'highest in database' to 'latest build id from buildbot'.
4. If anywhere in between, the process fails- it breaks and moves on to next builder.

mojca:
It seems more consistent now. Feel free to squash the first and third commit and merge.

mojca:
I don't know if there's a straightforward way to write tests to cover such scenarios. There definitely is, but maybe not a super easy one.

PR #89, opened 2019-08-15
Update the page '/statistics'

arjunsalyan:


arjunsalyan:
@mojca I have deployed these changes to EC2:

**General Stats**: http://ec2-52-34-234-111.us-west-2.compute.amazonaws.com/statistics/
 - Added 'days_ago' selector
 - Added duration and users count display
 - Added clt version chart

**Port Stats**: http://ec2-52-34-234-111.us-west-2.compute.amazonaws.com/port/wget/?tab=stats&days=30&days_ago=0
  - Added count of users to the duration display

arjunsalyan:
I believe saying just "X Users" feels a bit misleading.

mojca:
> I believe saying just "X Users" feels a bit misleading.

What about squeezing that information into the following sentence?
> Displaying statistics for X users who made submissions during Y days until Z days ago.

(I would say that in this case `Days` and `Days Ago` would need to be in lower case.)

app/ports/views.py
```diff
@@ -279,6 +279,7 @@ def stats(request):
     submissions_last_x_days = Submission.objects.filter(timestamp__range=[start_date, end_date]).order_by('user', '-timestamp').distinct('user')
     submissions_unique = Submission.objects.filter(id__in=Subquery(submissions_last_x_days.values('id')))
     macports_distribution = submissions_unique.values('macports_version').annotate(num=Count('macports_version'))
+    os_version_and_clt_version = sort_list_of_dicts_by_version(list(submissions_unique.values('clt_version', 'os_version').annotate(num=Count('user_id', distinct=True))), 'os_version')
```
mojca, line 283: Somehow the versions end up sorted as `9, 8, 6, 11, 10` ...

app/ports/views.py
```diff
@@ -252,12 +252,20 @@ def all_builds_filter(request):
 
 def stats(request):
     days = request.GET.get('days', 30)
+    days_ago = request.GET.get('days_ago', 0)
+
+    # Validate days and days_ago
+    for value in days, days_ago:
+        check, message = validate_stats_days(value)
```
mojca, line 260: Does this mean that duration of 0 days is valid?

(I don't care that much though ... I wouldn't mind if this wasn't so strict anyway.)

app/ports/templates/ports/stats.html
```diff
@@ -156,14 +156,28 @@
...
     <br>
     {% csrf_token %}
-    <form class="form-inline justify-content-center" method="get" action="{% url 'stats_home' %}#os-chart">
-        <label for="days"><h5>The charts below show stats for users who made submissions in last: </h5></label>
-        <select onchange="this.form.submit()" class="form-control ml-2" name="days" id="days">
-        {% for i in allowed_days|slice:"1:" %}
-            <option class="format-option" value="{{ i }}" {% if days == i %}selected{% endif %}>{{ i }} Days</option>
-        {% endfor %}
+    <form id="duration-selector" class="form-inline justify-content-center" method="get" action="{% url 'stats_home' %}#duration-selector">
+        <lable for="days">Displaying statistics for users who made submissions in previous:</lable>
```
mojca: Now this template looks the same as the one in `templates/ports/port-detail/installation_stats.html`. Could we use the exact same template (same file, not just same contents)?

mojca (commented):
Look ok. Some minor comments / suggestions only.

app/ports/views.py
arjunsalyan: I am sorry, which sorting are we referring to, OS Versions or CLT Versions? For me, both are sorted very well.

app/ports/views.py
arjunsalyan: Yeah, but the selectors in the template don't allow users to actually choose `0`.

app/ports/templates/ports/stats.html
arjunsalyan: This is quite different from that one. Charts like, `Users vs Month`, `Versions vs Month` and a lot of other things are different.

arjunsalyan:
> > Displaying statistics for X users who made submissions during Y days until Z days ago.

I like this one, I will change the page to use this.

arjunsalyan:
Done the change and is deployed here: http://ec2-52-34-234-111.us-west-2.compute.amazonaws.com/statistics/

app/ports/views.py
mojca: Clt versions are not sorted correctly for me on http://ec2-52-34-234-111.us-west-2.compute.amazonaws.com/statistics/ for example.

arjunsalyan:
This seems ready to merged?

mojca:
We can merge it and apply other minor changes later.

PR #95, opened 2019-08-18
git_update.py: Do not choke on non-port edits

neverpanic:
Commits such as [e2a40231da0ce1040195081f9e5375c23af07c75][1], which change files in the root directory of the repository caused this code to crash, because `'.travis.yml'.split('/')` does not have two elements. Avoid this by explicitly checking for the number of split elements.

```
  File "/code/app/manage.py", line 15, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.6/dist-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.6/dist-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.6/dist-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.6/dist-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
  File "/code/app/ports/management/commands/update-portinfo.py", line 33, in handle
    updated_portdirs = git_update.get_list_of_changed_ports(new_commit, options['old_commit'])
  File "/code/app/parsing_scripts/git_update.py", line 75, in get_list_of_changed_ports
    portdir = sections[0].lower() + '/' + sections[1].lower()
IndexError: list index out of range
```

Additionally, use a set rather than abusing a dict as return value and adjust the tests as necessary.

[1]: https://github.com/macports/macports-ports/commit/e2a40231da0ce1040195081f9e5375c23af07c75

### DIFF
--- a/app/parsing_scripts/git_update.py
+++ b/app/parsing_scripts/git_update.py
@@ -69,7 +69,7 @@ def get_list_of_changed_ports(new_commit=None, old_commit=None):
         changed_paths = subprocess.run([GIT, 'diff', '--name-only', range], stdout=subprocess.PIPE).stdout.decode(
             'utf-8')
         s = io.StringIO(changed_paths)
-        updated_ports_dir = {}
+        updated_ports_dir = set()
         for line in s:
             sections = line.split('/')
             if len(sections) < 2:
@@ -77,7 +77,7 @@ def get_list_of_changed_ports(new_commit=None, old_commit=None):
                 continue
             portdir = sections[0].lower() + '/' + sections[1].lower()
             if portdir not in updated_ports_dir:
-                updated_ports_dir[portdir] = ""
+                updated_ports_dir.add(portdir)
         os.chdir(BASE_DIR)
         print(updated_ports_dir)
         return updated_ports_dir
--- a/app/parsing_scripts/git_update.py
+++ b/app/parsing_scripts/git_update.py
@@ -72,6 +72,9 @@ def get_list_of_changed_ports(new_commit=None, old_commit=None):
         updated_ports_dir = {}
         for line in s:
             sections = line.split('/')
+            if len(sections) < 2:
+                # ignore updates in the root directory
+                continue
             portdir = sections[0].lower() + '/' + sections[1].lower()
             if portdir not in updated_ports_dir:
                 updated_ports_dir[portdir] = ""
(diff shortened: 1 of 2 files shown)


arjunsalyan (approved):

arjunsalyan:
Thank you very much, this looks much better.

mojca (approved):
Thank you very much for catching and fixing this.

Looks good to me, but we definitely need to add another test (to check that we don't break precisely this in the future), which compares, say 428a6a7dfe09452ce9c8f7c2c757d191998eaa13 .. e2a40231da0ce1040195081f9e5375c23af07c75 (two subsequent commits, one of which changes just travis config).

Can we also add a test which checks modfications of `_resources/macports1.0/xcode_versions.ini`? (To make sure that we don't end up trying to update information about port `macports1.0`, for example.)

Do we need any additional checks to ensure that neither of the first two elements starts with a dot or underscore? Or that it contains just lowercase letters `a-z`?

@arjunsalyan: can you please add the additional test(s)?

(I don't care if this gets merged immediately and additional tests are added later, or if the fixes are done before merging.)

neverpanic:
I thought about adding the test, but I'd rather leave writing the test up to you. Ideally, the test wouldn't need a checkout of the macports repository (and thus internet access) either.

I'd vote for merging this now, as updating the site is currently broken due to this.

arjunsalyan:
Yes, we should merge this and I will add related tests by the end of the day.

arjunsalyan:
> Can we also add a test which checks modfications of `_resources/macports1.0/xcode_versions.ini`? (To make sure that we don't end up trying to update information about port `macports1.0`, for example.)

The only place where the code could break was when splitting the changed path resulted in less than 2 elements. After this the code only tries to update those changed paths which are also available in PortIndex under `portdir`.

The code takes the set of updated paths and loops over port objects to check if the `portdir` is present in the set, if yes, it adds that `portdir` to the list that is sent to `Port.update()`.

> Do we need any additional checks to ensure that neither of the first two elements starts with a dot or underscore? Or that it contains just lowercase letters `a-z`?

As mentioned, it won't try to update those paths which are not available in the PortIndex. I can try writing tests for such situations.

> @arjunsalyan: can you please add the additional test(s)?

Yes, I will add them, thanks.

PR #96, opened 2019-08-19
Add test for non-port paths

arjunsalyan:


app/ports/tests/test_git_update.py
```diff
@@ -23,6 +23,12 @@ def test_between_new_and_database(self):
         ports = git_update.get_list_of_changed_ports("f3527caa506b1b944f43f47085dd35a8b8e2b050")
         self.assertEquals(ports, set(['net/nomad', 'sysutils/terraform']))
 
+    def test_non_port_paths(self):
+        ports = git_update.get_list_of_changed_ports("29129a78bafafe5b85dca4687707ce17223a355d",
+                                                     "e2a40231da0ce1040195081f9e5375c23af07c75"
```
mojca: Is the commit `e2a40231da0ce1040195081f9e5375c23af07c75` actually inclusive here? Doesn't this mean "anything that came after e2a40..."?

arjunsalyan: We are using `^..` to generate the range and as far I know includes the extremes. But I will move one commit behind to make this more reliable.

arjunsalyan: I moved one commit behind and it still works fine.

mojca: So this basically means that we are always updating the last port twice? We first update it once, write the commit of that change, and then update it again when new changes arrive?

The test is fine, and the behaviour cannot hurt, it just might not be the most optimal one.

PR #97, opened 2019-08-20
Simplify git_update.py script

arjunsalyan:
I felt this while writing the documentation that the function `git_update.get_list_of_changed_ports()` has needlessly become a bit complex.

This function should just take a new commit hash and an old commit hash and return the changed paths for the range of commits. It should not use any brain of its own in finding the hash if not provided. The work of generating old and new commits should be done outside of the function. For example, now it is being done by the command, `update-portinfo`.


umeshksingla:
I'll get back on this soon.

arjunsalyan:
This script has been completely rewritten and improved to take care of the previous concerns.

PR #98, opened 2019-08-20
Switch to lowercase to spell out 'Today'

mojca:
I still believe that we should be using a single template for duration selection (it could be a template that covers just a tiny part of the website, it doesn't need to actually show the graphs), so that this would reduce to fixing a single file, but I just wanted to fix this in the middle of a sentence.


arjunsalyan:
We can merge this, I will try to implement the template.

arjunsalyan (approved):

PR #100, opened 2019-08-22
Fix sorting of versions in CLT version chart

arjunsalyan:
I have deployed these changes to: http://ec2-52-34-234-111.us-west-2.compute.amazonaws.com/statistics/

Thank you @mojca for picking this out.

### DIFF
--- a/app/ports/templatetags/sort_version.py
+++ b/app/ports/templatetags/sort_version.py
@@ -1,3 +1,5 @@
+from distutils.version import LooseVersion
+
 from django import template
 
 register = template.Library()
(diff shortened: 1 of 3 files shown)


app/ports/templatetags/sort_version.py
```diff
@@ -5,4 +5,4 @@
 
 @register.filter('sortversion')
 def sortversion(distribution, key):
-    return sorted(distribution, key=lambda x: ((0, ) if x[key] == "none" else tuple(int(i) for i in x[key].split("."))))
+    return sorted(distribution, key=lambda x: ((0, 1) if x[key] == "" else (0, 2) if x[key] == "none" else tuple(int(i) for i in x[key].split("."))))
```
umeshksingla: Is it not possible to use the `LooseVersion` logic here as well, similar to port utilities?

arjunsalyan: Oh, right. Thanks. Got missed by me totally.

PR #103, opened 2019-08-24
Port detail: Add message for deleted ports

arjunsalyan:


app/ports/templates/ports/portdetail.html
```diff
@@ -34,6 +34,15 @@
     {% endif %}
     <!-- Obsolete port warning end-->
 
+    <!-- Deleted port warning start-->
+    {% if req_port.active is False %}
+        <div class="alert alert-danger" role="alert">
+            This port has been deleted. This page will exist till another port with the name "{{ req_port.name }}" comes
```
mojca, line 40: I support merging this immediately, but here's another somewhat weird suggestion / request for which I would like to get some feedback.

When someone is looking for a particular (deleted) port, there might be a great chance that this person actually wants to install the port that everyone else thought was useless, or at least broken beyond repair. Maybe that person is even wiling to look into fixing the old `Portfile`.

What we are missing in MacPorts is precisely the way to easily find old deleted ports, and to find reasons why they were deleted. We have almost all functionality in place now to allow us to store the commit shasum in which the port was deleted, so we could provide URL to the commit on github, which would hopefully include the commit message explaining why the port was deleted, potentially also showing the link to the pull request with the relevant discussion. And more importantly, it would be easy to see what the contents of the old port were.

What do you think about adding either an additional column (or maybe an additional table since the number of entries is expected to be relatively low) listing `(port_name, commit, <timestamp?>)` for deleted ports?

As an example the port [maven-ant-tasks](https://ports.macports.org/port/maven-ant-tasks/) could link to https://github.com/macports/macports-ports/commit/493ec90efd3be0db81ca988968486eecb5b58fdd.

@cjones051073 @ryandesign @neverpanic

mojca: Also, we probably need to remove `View on GitHub` link for such ports.

arjunsalyan: I don't think that it would be very easy to find the commit in which the port was deleted. We just get a list to changed paths. And it might become complex when one subport is deleted and one is updated- we won't be able to tell in which commit update took place and in which deletion. Please correct me, if I am missing something.

How about simply displaying a link to such a query:
https://github.com/search?q=maven-ant-tasks+delete+repo:macports/macports-ports&type=Commits

It is definitely not very reliable, but can at least give a direction.

PR #109, opened 2019-09-05
Fix errors in detecting deleted ports

arjunsalyan:
I still need to perform more tests. And look for any other situations that are being missed, if any. 

Currently, this mechanism to detect deleted ports is totally broken:
- When the main port is removed (portdir is removed), it is not being marked as deleted.
- If one of the subports is deleted, all other subports and parent ports is being marked as deleted.

This PR should fix the above problems.

### DIFF
--- a/app/ports/management/commands/update-portinfo.py
+++ b/app/ports/management/commands/update-portinfo.py
@@ -32,20 +32,24 @@ class Command(BaseCommand):
         # The code will then use the commit from last update which is stored in the database
         updated_portdirs = git_update.get_list_of_changed_ports(new_commit, options['old_commit'])
 
-        # Using the received list of port names, find related JSON objects
+        # Generate a dictionary containing all the portdirs and initialise their values
+        # with empty sets. The set would contain the ports under that portdir.
+        dict_of_portdirs_with_ports = {}
+        for portdir in updated_portdirs:
+            dict_of_portdirs_with_ports[portdir] = set()
+
+        # Using the received set of updated portdirs, find corresponding JSON objects for all ports under
+        # that portdir.
         ports_to_be_updated_json = []
-        found_ports_tree = {}
         for port in data['ports']:
             portdir = port['portdir'].lower()
             portname = port['name'].lower()
             if portdir in updated_portdirs:
                 ports_to_be_updated_json.append(port)
-                if found_ports_tree.get(portdir) is None:
-                    found_ports_tree[portdir] = {}
-                found_ports_tree[portdir][portname] = True
+                dict_of_portdirs_with_ports[portdir].add(portname)
 
         # Mark deleted ports
-        Port.mark_deleted(found_ports_tree)
+        Port.mark_deleted(dict_of_portdirs_with_ports)
 
         # Run updates
         Port.update(ports_to_be_updated_json)
--- a/app/ports/models/port.py
+++ b/app/ports/models/port.py
@@ -276,10 +276,10 @@ class Port(models.Model):
         # ============ END ==============
 
     @classmethod
-    def mark_deleted(cls, dict_of_portdirs):
-        for portdir in dict_of_portdirs:
+    def mark_deleted(cls, dict_of_portdirs_with_ports):
+        for portdir in dict_of_portdirs_with_ports:
             for port in Port.objects.filter(portdir__iexact=portdir).only('portdir', 'name', 'active'):
-                if port.name not in dict_of_portdirs[portdir]:
+                if port.name not in dict_of_portdirs_with_ports[portdir]:
                     port.active = False
                     port.save()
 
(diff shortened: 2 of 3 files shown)


arjunsalyan:
http://ec2-52-34-234-111.us-west-2.compute.amazonaws.com/port/maven-ant-tasks/summary/

Worked for the previous cases which were failing.

arjunsalyan:
@mojca @umeshksingla 

I would like to bring our attention to this serious bug. I am sorry, I was busy with exams and could not take this as a priority earlier.

app/ports/management/commands/update-portinfo.py
```diff
@@ -44,6 +44,12 @@ def handle(self, *args, **options):
                     found_ports_tree[portdir] = {}
                 found_ports_tree[portdir][portname] = True
 
+        # When entire portdir is removed, it won't be picked up while looping over portindex.json
```
mojca: I find the name of function `handle` somewhat non-intuitive, as well as its description somewhat not reflecting the exact purpose.

Could we replace `found_ports_tree` in this function and `dict_of_portdirs` inside `mark_deleted` with a variable using the same name at both places? Maybe `dict_of_portdirs_with_ports` (`dict_of_portdirs_with_set_of_ports`). That's not the best name either, but maybe you can think of something better.

app/ports/management/commands/update-portinfo.py
```diff
@@ -44,6 +44,12 @@ def handle(self, *args, **options):
                     found_ports_tree[portdir] = {}
                 found_ports_tree[portdir][portname] = True
```
mojca: This could probably be a `set` rather than dictionary?

app/ports/management/commands/update-portinfo.py
```diff
@@ -44,6 +44,12 @@ def handle(self, *args, **options):
...
 
+        # When entire portdir is removed, it won't be picked up while looping over portindex.json
+        # Add it by looping over update_portdirs again
+        for portdir in updated_portdirs:
+            if portdir not in found_ports_tree:
+                found_ports_tree[portdir] = {}
```
mojca: Can we move this loop on top and just keep
```python
for portdir in updated_portdirs:
    dict_of_portdirs_with_ports[portdir] = {} # or maybe better a set
```
This way you will first create a list of portdirs which need attention and only then populate them as needed with the above loop.

mojca (commented):
Other than what I wrote in comment this looks good to me. I leave it up to you to change as you find suitable and merge it.

But please add at least another unit test (either now or later; adjust names as you see fit):
* you have port `A_main` and its subports `A_sub1`, `A_sub2`, `A_sub3`
* `A_main` gets deleted, but `A_sub1` is moved to its own path together with `A_sub2`
You need to ensure that you don't mark the ports `A_sub1` and `A_sub2` as deleted.

And another unit test where everything gets deleted.

app/ports/management/commands/update-portinfo.py
mojca: If I'm missing something about why this wouldn't work, please let me know.

arjunsalyan: Yeah, that would be a lot better. Thanks.

mojca (approved):
Looks good.

arjunsalyan:
The ports which were deleted since the app got live are incorrectly marked as active in app's database. We either need to make a long update run (since the commit made at the time of deploying the app) or we can make a new command just to take care of marking the ports as deleted.

It would need running a command on the server.

@mojca @neverpanic

pmetzger:
How long would a "long update run" take?

mojca:
We should probably at some point bite the bullet and occasionally run a full update (not between two commits, but simply updating all ports to the latest version, just in case). But this only makes sense with a fully optimized version.

Now, such an action would need to be composed out of two parts:
- get all the existing ports up to date
- mark all non-existing ports as deleted

and maybe we could implement the second one anyway.

I'm thinking about the following, but please correct me in case it doesn't make sense.

Maybe we could once per day or every few days run the following (as a transaction, if possible, and making sure that we don't run any other update in the meantime):
* read the json file (and make sure that we end up with a hash with port names)
* fetch all the active ports from the database (just the columns `name` and `active`)
* iterate through the list of all results, check for every one of them if the port name exists in the hash from json; if not, mark the port as inactive, else continue
* save all changes in a single database transaction

The important part would be not to update json in the meantime or start updating other ports in the database.

This action should be pretty fast and is generic enough to justify simply running it every now and then. Ideally we would write it in such a way that nobody would have to actually touch the server to set up new cron jobs.

arjunsalyan:
This looks right, so what we can do is add this function to the already running cronjob `update-portinfo` and make it run only once a week, maybe on some given day?

pmetzger:
If the run only takes a few minutes, why not do it daily?

PR #111, opened 2019-09-07
Allow "." in variant page url

arjunsalyan:


app/MacPorts/urls.py
```diff
@@ -35,7 +35,7 @@
     path('ports/filter/variant/', views.search_ports_in_variant, name='search_ports_in_variant'),
     path('ports/load_tickets/', views.tickets, name='trac_tickets'),
     path('ports/category/<slug:cat>/', views.categorylist, name='category_list'),
-    path('ports/variant/<slug:variant>/', views.variantlist, name='variant_list'),
+    url(r'^ports/variant/(?P<variant>[-a-zA-Z0-9_.]+)/$', views.variantlist, name='variant_list'),
```
jmroot: If this is intended to only match valid variant names, then the dash character shouldn't be in there.

arjunsalyan: I don't know, having a bit flexible url might be fine?

mojca: Honestly I wasn't even aware that dots became allowed in variant names (I was using underscores in the past, for example for `+perl5_28`, where a dot would make a lot more sense; I'm not sure if I want to change any of those now though: too much work for too little gain).

I kind of agree that having a bit more flexibility cannot hurt, I just want to point out that both `+` and `-` are explicitly forbidden even if we add more characters to the allowed set in the future, since these two characters usually represent delimiters between variants; in the same way as you cannot have `+` or `-` as valid character in variable names in `C` or `python`.

Again, flexibility cannot hurt, I wouldn't mind if this was part of some default allowed character set, but in case when you are explicitly listing the allowed characters, it might actively confuse future developers trying to figure out why a forbidden character was on the list in the first place. For this reason I would tend to agree with @jmroot.

My suggestion is to remove the dash and merge this. It might be nice to additionally (later) add some minor test case (broken before this commit and passing after), even though I'm not sure what that would be. I wouldn't wait for the test case, this needs to be merged ASAP as it's fixing a "seriously broken" behaviour.

mojca (approved):

umeshksingla:
Okay, I’d also remove the dash and then merge. On the other hand, should we
have this type of validation while inserting?

app/MacPorts/urls.py
arjunsalyan: Okay, thank you. I have merged this after removing the dash.

mojca:
> On the other hand, should we have this type of validation while inserting?

I cannot really decide. The data is computer-generated, so we don't generally need so strict validation. And if we do add validation: how would you handle the failures in updates? If we would add validation, it would probably be best on the database level, but I don't consider it super necessary.

(In principle almost any unicode string could be made to work if properly encoded in the URL, so I fail to see too many benefits.)

Issue #114, opened 2019-09-21
Proposal: drop dependency on Google Charts


l2dy:
https://github.com/macports/macports-webapp/blob/b753285bca2771cb8bb873e0030345df3defeab8/app/ports/templates/ports/portdetail.html#L9

As stated in the [Google Charts FAQ](https://developers.google.com/chart/interactive/faq), their ToS disallows "offline" usage. Which has the following effects:
1. This website has a hard dependency on Google, and must adapt if they made backwards incompatible changes.
2. Visits to these pages could be tracked by Google.
3. In certain restricted networks where Google's APIs are unavailable, some pages would be unusable.

mojca:
It's a pitty that this wasn't brought to attention slightly earlier as we could use a different api from the start.

I assume you saw it blocked yourself?

Which graphing tool would you suggest to use. There are plenty easy to use libraries around, I don't have much experience with either.

l2dy:
[D3](https://github.com/d3/d3) is a very powerful JavaScript library for visualizing data. We could use a charting library based on it, e.g. [c3](https://github.com/c3js/c3) or [plotly.js](https://github.com/plotly/plotly.js).

umeshksingla:
[HIghCharts](https://www.highcharts.com/) is an excellent alternative, I'd like if you can take a look at its [Non-Commercial FAQ](https://shop.highsoft.com/faq#Non-Commercial-0) because I did not get it fully. It allows [downloading the scripts](https://www.highcharts.com/blog/download/) for offline usage.

umeshksingla:
> [D3](https://github.com/d3/d3) is a very powerful JavaScript library for visualizing data. We could use a charting library based on it, e.g. [c3](https://github.com/c3js/c3) or [plotly.js](https://github.com/plotly/plotly.js).

Yes, this. I have definitely heard about it a lot but have no experience with it.

l2dy:
> [HIghCharts](https://www.highcharts.com/) is an excellent alternative, I'd like if you can take a look at its [Non-Commercial FAQ](https://shop.highsoft.com/faq#Non-Commercial-0) because I did not get it fully. It allows [downloading the scripts](https://www.highcharts.com/blog/download/) for offline usage.

"The license allows for up to 5 developers to work on our software."

Nope, don't use this.

---

TBH, I prefer depending on open source software only. And slightly OT, we should add a LICENSE file to this repository.

mojca:
[...]

PR #115, opened 2019-09-21
Add BSD license

mojca:
See #114.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,25 @@
+BSD 2-Clause License
+
+Copyright (c) 2019, The MacPorts Project
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+1. Redistributions of source code must retain the above copyright notice, this
+   list of conditions and the following disclaimer.
+
+2. Redistributions in binary form must reproduce the above copyright notice,
+   this list of conditions and the following disclaimer in the documentation
+   and/or other materials provided with the distribution.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2019 MacPorts
```
l2dy: Should we use ", The MacPorts Project." here? It's used in macports-base and ports.

l2dy:
cc @macports/portmgr 

Is other permissive license like the MIT license OK for new projects, or should we use the same BSD-style license across all official MacPorts repositories?

LICENSE
mojca: I actually wanted to ask because I wasn't sure what to use as the copyright holder. You are probably right that `The MacPorts Project` might make more sense, in particular if used in other repos.

mojca (commented):
> Is other permissive license like the MIT license OK for new projects, or should we use the same BSD-style license across all official MacPorts repositories?

I'm in fact interested in opinion from other developers, but eventually nearly all the code so far has been written by @arjunsalyan (it would be mainly up to him to decide if he had a strong opinion, even though ... it would be up to us to provide some "licence restrictions" before he even started working on the project :).

In particular we don't want to have the same licence mentioning Apple etc., but pure BSD sounds fine to me as well.

(I use MIT in another project which requires a lot of interoperability with projects licenced under various other licences, and this one sounded like the least common denominator to all. So in a way a "very very permissive licence")

mojca:
I switched to 2-clause BSD as suggested by @neverpanic and @jmroot (using the default template provided by GitHub). I guess this is good to go then?

l2dy:
> I switched to 2-clause BSD as suggested by @neverpanic and @jmroot (using the default template provided by GitHub). I guess this is good to go then?

All contributors need to agree before we can apply this license.

mojca:
Sure, but apart from a trivial typo fix from @amake that's basically just @arjunsalyan (along with @umeshksingla and @neverpanic for a few other minor changes), so I would just wait for his final confirmation.

This reminds me on my inability to fix a trivial bug in hyphenation patterns for OpenOffice since I didn't want to bother about signing & mailing them a fully fledged contract. Does dependabot need to agree as well? :) :) :)

arjunsalyan (approved):

arjunsalyan:
Thank you very much @mojca . This looks good to me.

umeshksingla (approved):

PR #116, opened 2019-09-21
Fix x-axis sorting in port versions chart

### LINKED ISSUE #112: Strange x axis sorting for installed port versions chart
Sorting of the port versions in the chart showing the installed versions appears to be done in decreasing order of frequency, so the version with the most stats on the left, least on the right. In my opinion this is a strange inconvenient scheme. It would be better to sort accordingly to the version itself, so oldest in the left, newest in the right.

See attached screen shot taken from

https://ports.macports.org/port/clang-8.0/stats?days=30&days_ago=0

PNG image

My suggestion would flip 8.0.0_0 with 8.0.0_1

arjunsalyan:
Closes: https://github.com/macports/macports-webapp/issues/112

### DIFF
--- a/app/ports/views.py
+++ b/app/ports/views.py
@@ -186,7 +186,7 @@ def portdetail_stats(request):
     port_installations = PortInstallation.objects.filter(submission_id__in=Subquery(submissions.values('id')), port__iexact=port_name)
     requested_port_installations_count = port_installations.filter(requested=True).aggregate(Count('submission__user_id', distinct=True))
     total_port_installations_count = port_installations.aggregate(Count('submission__user_id', distinct=True))
-    port_installations_by_port_version = port_installations.values('version').annotate(num=Count('version')).order_by('-num')
+    port_installations_by_port_version = sort_list_of_dicts_by_version(list(port_installations.values('version').annotate(num=Count('version'))), 'version')
     port_installations_by_os_and_xcode_version = sort_list_of_dicts_by_version(list(port_installations.values('submission__xcode_version', 'submission__os_version').annotate(num=Count('submission__user_id', distinct=True))), 'submission__os_version')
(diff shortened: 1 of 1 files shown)


arjunsalyan:
This is a minor change, so I believe it can be merged right away.

mojca:
I'm on my phone at the moment. If you tested it, feel free to simply merge it.

cjones051073:
Just for my info, When will this fix hit the main server ? The issue is still present at

https://ports.macports.org/port/clang-8.0/stats?days=30&days_ago=0

arjunsalyan:
It has been deployed. The versions are sorted correctly for me. Highest version (8.0.1_0) on left and decreasing as we go right.

cjones051073:
> It has been deployed. The versions are sorted correctly for me. Highest version (8.0.1_0) on left and decreasing as we go right.

Thats the wrong way around, for me. Graphs have the lowest (oldest) on the left and highest(newest) on the right. What you have is backwards to that standard convention. 

Actually, now I look again I would argue all your graphs have the x axis sorting backwards...

mojca:
I think we started discussing at some time whether to sort the OS versions chronologically from left to right, or whether to start with the one that's most important on the left, and then you can mostly ignore the rest of the x axis.

As far as "User vs Version for each month" (which could get a better name) goes, I would leave the sorting on the y axis precisely as it is (the latest version on top). Same for xcode versions etc.

For port version & macports version I would tend to agree that increasing version from left to the right kind of makes more sense.

I'm pretty much indifferent about the sorting order on the x axis for macOS versions, but if we want to keep it consistent with the rest, it might as well be swapped around. But we'll probably need to remove `OS:` prefix which is by now taking way too much space.

I don't have super strong feelings though.

PR #118, opened 2019-09-24
Add test for moved ports

arjunsalyan:
This is to support the bug-fix made in https://github.com/macports/macports-webapp/pull/109 .

### DIFF
--- a/app/ports/tests/test_update_portinfo.py
+++ b/app/ports/tests/test_update_portinfo.py
@@ -29,7 +29,137 @@ class TestUpdatePortinfo(TransactionTestCase):
         self.assertEquals(port_status_mainport, True)
         self.assertEquals(port_status_subport, False)
 
+    def test_moved(self):
+        Port.update([
+            {
+                "variants": ["universal"],
+                "portdir": "categoryA/port-A1",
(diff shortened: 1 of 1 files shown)


app/ports/tests/test_update_portinfo.py
```diff
@@ -29,7 +29,137 @@ def test_deleted(self):
...
+        Port.update([
+            {
+                "variants": ["universal"],
+                "portdir": "categoryA/port-A1",
+                "depends_fetch": ["bin:port-A4:port-A4"],
+                "description": "This is port A1 of categoryA",
+                "homepage": "http:\/\/portA1.website\/",
+                "epoch": "0",
+                "platforms": "darwin",
+                "name": "port-A2-subport",
```
mojca, line 42: Just to clarify it for myself: in what scenario do you end up calling `Port.update` with just the list of two subports in the main port's path?

arjunsalyan: I am using this to add two more subports to the main port 'port-A1', ideally all ports under that portdir should get updated, but I am using this as a workaround. This happens only for this test.

Issue #117, opened 2019-09-24
Incorporate the new builders


mojca:
According to @ryandesign the old builders will stay offline for a while in order to be able to access the logs, while the new builders (dropping the `_legacy` suffix) will soon get online, building binaries against the `libc++` library, so we will never get two builders running in parallel.

For that reason I expect the port health to show just a single entry per OS, showing the latest build on either the new "non-legacy" builder, or on the old "legacy" builder, whichever came last. Filter based on builder should probably also show just a single entry per OS/arch pair.

In a not too distant future we expect to switch to a new setup where builder names might change again (at the moment dots are forbidden in names, for example), and we would want to keep the relationship of a single OS/arch pointing to potentially multiple different builders.

It would be nice to prepare for this sooner than later, ideally before Ryan turns the builders back on.

At the moment we are using
```
(builder_name, build_id, port_name, status, ...)
```
for individual builds, where `builder_name` might be something like `10.6_i386_legacy`.

Maybe we need an additional column pointing to "10.6_i386", so that both old builder name `10.6_i386_legacy` and the new `10.6_i386` (as well as potentially future `10_6_i386`) would point to the same "logical unit" of a 32-bit 10.6 worker (either as a string or an integer pointing to a different table)? Maybe we would want to add some version as well? Ideas welcome.

@umeshksingla @neverpanic

jmroot:
> It would be nice to prepare for this sooner than later, ideally before Ryan turns the builders back on.

They've been on for a few hours FYI.

arjunsalyan:
We can add a new column `os_id` to the `Builder` table:

| pk | name | os_id |
----|------|------|
1 | 10.6_x86_64 | 1 |
2 | 10.6_x86_64_legacy | 1 |
3 | 10.6_i386_legacy | 2 | 
4 | 10.6_i386 | 2 |

And the `BuildHistory` table could remain same:

| builder_name | build_id | status | port_name | time_start | ... |
|----|----|----|----|---|----|
1 | 3456 | build successful | portA | 2019-09-20 19:03:57 | ... |

`builder_name` field is the foreign key to the `Builder` table.

And then we could simply use `os_id` instead of the currently being used `name` field to filter the build history related to that builder.

Are you suggesting the same?

mojca:
[...]

PR #119, opened 2019-09-25
Add non legacy builders (<10.9)

arjunsalyan:
I will soon be doing the table related changes, but till then it would be better to not miss on the build history for the new builders.

https://github.com/macports/macports-webapp/issues/117


arjunsalyan:
I have applied this to the test site:

http://ec2-52-34-234-111.us-west-2.compute.amazonaws.com/ports/all_builds/

mojca:
I would ideally like to see just one instance of 10.6 on port health status, and just one entry in the filter.

arjunsalyan:
> I would ideally like to see just one instance of 10.6 on port health status, and just one entry in the filter.

I am working on the new table design, shall I simply hardcode the `os_arch_key` and `display_name` columns?

arjunsalyan:
This PR is not in the stage of merging right now, as it will cause migration conflicts with https://github.com/macports/macports-webapp/pull/126 .

arjunsalyan:
I have deployed this on the test site. I am utilising only one additional field `display_name`.

Does it seem right now? @mojca 
http://ec2-52-34-234-111.us-west-2.compute.amazonaws.com

mojca:
Can you please try to import at least a few builds (even if just ten builds per worker)? I don't yet see any builds shown after the 22nd of August to be able to see how it works.

arjunsalyan:
> Can you please try to import at least a few builds (even if just ten builds per worker)? I don't yet see any builds shown after the 22nd of August to be able to see how it works.

I have started the process, the fetching should finish within next 10-15 minutes.

arjunsalyan:
Also, please do a hard refresh in the browser so that the changes to the javascript get reflected.

mojca:
Can you please point me to some actual builds?

arjunsalyan:
> Can you please point me to some actual builds?

On the page: http://ec2-52-34-234-111.us-west-2.compute.amazonaws.com/ports/all_builds/ select builder `10.8`.
Now, the first 20 builds refer to the builder `10.8_x86_64` (2285 to 2304) while the builds below it refer to the builder `10.8_x86_64_legacy`.

The app is initially set to fetch only 20 builds for those builders which do not have any builds in the database, that is why we only get the latest 20 builds for the newly added `10.8_x86_64`. We will need to decide this number.

And I also missed adding non-legacy builder for `10.6_i386`.

mojca:
Perfect, thank you. They are there now, but I didn't see them earlier.

The all_builds works precisely as desired. The port health on the other hand is picking up the legacy workers, maybe because you sorted by build number (but I didn't check).

That said, the port health has other flaws, like: the user doesn't see if the worker is offline, or the latest build is still pending (10.5/ppc has been offline for ages and port health shows just whatever was built last); I would attempt to fix this with buildbot 2.

Fetching just the latest 20 builds sounds fine. But we would need to distinguish between two different scenarios:
* users or temporary servers who just want to test a new feature (we don't want them to fetch all builds, at least not automatically)
* production server where we really want to fetch all builds (and we need to try to do this before they go "out of focus" and get deleted)

I'm ok with merging this now and fixing the issue with port health immediately after that.

app/ports/views.py
```diff
@@ -87,15 +87,15 @@ def portdetail(request, name, slug="summary"):
         if tab not in allowed_tabs:
             return HttpResponse("Invalid tab requested. Expected values: {}".format(allowed_tabs))
 
-        all_latest_builds = BuildHistory.objects.all().order_by('port_name', 'builder_name', '-build_id').distinct('port_name', 'builder_name')
-        port_latest_builds = list(BuildHistory.objects.filter(id__in=Subquery(all_latest_builds.values('id')), port_name__iexact=name).values('builder_name__name', 'build_id', 'status'))
+        all_latest_builds = BuildHistory.objects.all().order_by('port_name', 'builder_name__display_name', '-build_id').distinct('port_name', 'builder_name__display_name')
+        port_latest_builds = list(BuildHistory.objects.filter(id__in=Subquery(all_latest_builds.values('id')), port_name__iexact=name).values('builder_name__name', 'builder_name__display_name', 'build_id', 'status'))
```
mojca: We can no longer sort by build id. Maybe by `time_start`?

app/ports/migrations/0011_populate_display_name.py
```diff
@@ -0,0 +1,31 @@
+# Generated by Django 2.2.4 on 2019-10-09 14:19
+
+from django.db import migrations
+
+
+def add_new_builders(apps, schema_editor):
+    Builder = apps.get_model('ports', 'Builder')
+    Builder.objects.create(name='10.8_x86_64')
+    Builder.objects.create(name='10.7_x86_64')
+    Builder.objects.create(name='10.6_x86_64')
```
mojca, line 10: As already noted, we are missing the 10.6_i386 one.

mojca (approved):
Two hopefully minor changes to fix, and then we can merge it.

And then separately think of the best way to fetch all builds just on production machine.

app/ports/views.py
arjunsalyan: Yes, thank you.

arjunsalyan:
This is probably ready to be merged now? I am sorry for the delay.

mojca:
Cool, perfect.

Now we just need a way to fetch the old builds (fetch them all on the main server without having to fetch them on some test instances).

mojca:
Ok, we are still missing a few fixes. Looking at http://ec2-52-34-234-111.us-west-2.compute.amazonaws.com/port/qt59-qtserialport/, the builds are still picking the wrong builder. It works correctly if you switch to a different tab.

It might make sense to clean up the code a bit, but for now at least we should fix that bug.

arjunsalyan:
> Ok, we are still missing a few fixes. Looking at http://ec2-52-34-234-111.us-west-2.compute.amazonaws.com/port/qt59-qtserialport/, the builds are still picking the wrong builder. It works correctly if you switch to a different tab.

I am sorry, could you please specify it more? Also, the recent changes (made today) have not been deployed to the test site due to migration conflicts.

mojca:
> I am sorry, could you please specify it more?

Hmmm ... I can no longer reproduce what I saw. The most likely explanation is that I was accidentally looking/clicking at the 10.7 builder which still points to the legacy build. It might have to do with me travelling on a bumpy road while trying to concentrate on the screen :) :) :)

PR #121, opened 2019-09-25
Fix sorting of macports version on '/statistics'

arjunsalyan:


app/ports/views.py
```diff
@@ -464,7 +464,7 @@ def tickets(request):
     response = requests.get(URL)
     Soup = BeautifulSoup(response.content, 'html5lib')
     all_tickets = []
-    for row in Soup.findAll('tr', attrs={'class': ['color2-even', 'color2-odd']}):
+    for row in Soup.findAll('tr', attrs={'class': ['color2-even', 'color2-odd', 'color1-even', 'color1-odd']}):
```
mojca: This one must have ended up here by accident.

app/ports/views.py
```diff
@@ -283,7 +283,7 @@ def stats(request):
     # System Stats for Current Users
     submissions_last_x_days = Submission.objects.filter(timestamp__range=[start_date, end_date]).order_by('user', '-timestamp').distinct('user')
     submissions_unique = Submission.objects.filter(id__in=Subquery(submissions_last_x_days.values('id')))
-    macports_version = submissions_unique.values('macports_version').annotate(num=Count('macports_version'))
+    macports_version = sort_list_of_dicts_by_version(list(submissions_unique.values('macports_version').annotate(num=Count('macports_version'))), 'macports_version')
```
mojca, line 286: (Without testing) this probably works fine.

mojca (commented):
This can be merged once the accidental change is removed (even though I think that rebase will remove the accidental change anyway).

app/ports/views.py
arjunsalyan: Oops!

Issue #110, opened 2019-09-05
Don't display deleted ports unless the user asks you to


pmetzger:
Most of the time, if I'm searching MacPorts I'm looking for information about ports that are currently in the tree. It is relatively rare that I want to see data on deleted ports. So, I'd like to ask that unless the user checks a box or otherwise indicates they want to see deleted ports as well, that the deleted ports not be shown.

arjunsalyan:
Thank you, I wrote a comment relating to this in #108 .

@mojca

pmetzger:
Hi, just curious if this can be acted on? I'm finding it much harder working with (for example) the ocaml part of the tree where I got rid of a lot of obsolete ports because I can't filter the deleted ones out...

arjunsalyan:
I am working on it, there was a bug that prevented the deleted ports from getting detected, it just got fixed yesterday.
I will add the functionality to filter out deleted ports by this weekend.

pmetzger:
Thank you so much, @arjunsalyan!

arjunsalyan:
I have deployed a test version (a checkbox below the search bar):
http://ec2-52-34-234-111.us-west-2.compute.amazonaws.com/

Could you please see if it works fine and satisfies the requirement?
You may need to clear website data or do a hard refresh before it works as expected!

pmetzger:
Seems good to me...

mojca:
See also #283.

mascguy:
I definitely would prefer that deleted ports be hidden by default, with the current checkbox changed to "Show deleted ports." (Disabled by default.)

Indeed, checking the box to hide deleted ports is a necessity for me personally, as there are a boatload of openmpi-* and mpich-* subports that were removed last year.

Just curious, why was this closed? Is the intention to implement this as part of #283?

EDIT: Never Mind, inverting this is indeed covered by 283. Sorry folks!

Issue #44, opened 2019-07-20
Format of the License field


umeshksingla:
Not a super important issue but the license field can be improved sometime in the future to have a consistent format and without braces on the ends.
1. https://ports.macports.org/port/automake
2. https://ports.macports.org/port/aqbanking5

mojca:
However I would like to ask the experts for some hints first. I believe there is an actual difference when braces are used. In some cases it may mean "either one licence or the other", in others it may mean "both licences apply".

jmroot:
https://github.com/macports/macports-base/blob/master/src/port/port.tcl#L2186

So `MIT {PSF ZPL} {Apache-2 BSD}` is displayed in more human-readable form as `MIT and (PSF or ZPL) and (Apache-2 or BSD)`.

mojca:
@jmroot: thank you. I never noticed that it's formatted in this way when you use `port info`.

I have two questions though:
* How could we use this function directly to change what gets written to `portindex.tcl`? (I assume we can copy-paste the code, but it would be nicer to have just one reusable function called multiple times.)
* Why the following discrepancy?
```
> port info automake
automake @1.16.1_1 (devel)

Description:          Automake is a tool for automatically generating Makefile.in files from files called Makefile.am. Each Makefile.am is basically a series of
                      make variable definitions, with rules being thrown in occasionally. The generated Makefile.in files are compliant with the GNU Makefile
                      standards.
Homepage:             https://www.gnu.org/software/automake

Build Dependencies:   autoconf
Platforms:            darwin, freebsd
License:              (GPL-2+ or Permissive)
Maintainers:          none
```

```
> port info --license automake
license: {GPL-2+ Permissive}
```

My suggestion would be to change this in `portindex.tcl` and don't do anything in the app.

jmroot:
We need it to be machine-readable in the PortIndex.

mojca:
Sorry, I meant `PortIndex.json`, rather than `PortIndex` itself. (I made a typo in the file extension.)
Proper machine-readability in the scope of `json` would likely mean encoding that information in `json` constructs, but that would likely be an overkill. In itself (when forgetting `tcl`) the format with parenthesis and using `and`/`or` is probably equally machine-readable if one would want to do something further with "`MIT and (PSF or ZPL) and (Apache-2 or BSD)`" than when using curly braces?

PR #122, opened 2019-09-26
Allow filtering out deleted ports and loop over all ports to identify deleted ports

arjunsalyan:
I ran the tests on EC2 instance and the process was really fast, finished within 10-15 seconds.

So, I believe we can run it with every update and get rid of the old confusing method to detect deleted ports altogether.

### DIFF
--- a/app/ports/filters.py
+++ b/app/ports/filters.py
@@ -18,4 +18,4 @@ class PortFilterByMultiple(django_filters.FilterSet):
 
     class Meta:
         model = Port
-        fields = ['categories__name', 'maintainers__name', 'maintainers__github']
+        fields = ['categories__name', 'maintainers__name', 'maintainers__github', 'active']
--- a/app/ports/management/commands/update-portinfo.py
+++ b/app/ports/management/commands/update-portinfo.py
@@ -32,11 +32,7 @@ class Command(BaseCommand):
         # The code will then use the commit from last update which is stored in the database
         updated_portdirs = git_update.get_list_of_changed_ports(new_commit, options['old_commit'])
 
-        # Generate a dictionary containing all the portdirs and initialise their values
-        # with empty sets. The set would contain the ports under that portdir.
-        dict_of_portdirs_with_ports = {}
-        for portdir in updated_portdirs:
-            dict_of_portdirs_with_ports[portdir] = set()
+        all_ports_from_json = set()
 
         # Using the received set of updated portdirs, find corresponding JSON objects for all ports under
         # that portdir.
@@ -44,15 +40,17 @@ class Command(BaseCommand):
         for port in data['ports']:
             portdir = port['portdir'].lower()
             portname = port['name'].lower()
+            all_ports_from_json.add(portname)
             if portdir in updated_portdirs:
                 ports_to_be_updated_json.append(port)
-                dict_of_portdirs_with_ports[portdir].add(portname)
-
-        # Mark deleted ports
-        Port.mark_deleted(dict_of_portdirs_with_ports)
 
         # Run updates
         Port.update(ports_to_be_updated_json)
 
         # Write the commit hash into database
         LastPortIndexUpdate.update_or_create_first_object(data['info']['commit'])
+
+        # Make a thorough run to check for deleted ports
+        print("Starting the thorough run.")
+        Port.mark_deleted(all_ports_from_json)
+        print("Thorough Run finished.")
--- a/app/ports/models/port.py
+++ b/app/ports/models/port.py
@@ -16,6 +16,11 @@ class PortManager(models.Manager):
         return self.get(name=name)
 
 
+class ActivePortsManager(models.Manager):
+    def get_queryset(self):
+        return super(ActivePortsManager, self).get_queryset().filter(active=True)
+
+
 class Port(models.Model):
     portdir = models.CharField(max_length=100)
     description = models.TextField(default='')
@@ -33,6 +38,7 @@ class Port(models.Model):
     active = models.BooleanField(default=True)
 
     objects = PortManager()
+    get_active = ActivePortsManager()
 
     @classmethod
     def load(cls, data):
--- a/app/ports/models/port.py
+++ b/app/ports/models/port.py
@@ -276,10 +276,10 @@ class Port(models.Model):
         # ============ END ==============
 
     @classmethod
-    def mark_deleted(cls, dict_of_portdirs_with_ports):
-        for portdir in dict_of_portdirs_with_ports:
-            for port in Port.objects.filter(portdir__iexact=portdir).only('portdir', 'name', 'active'):
-                if port.name not in dict_of_portdirs_with_ports[portdir]:
+    def mark_deleted(cls, all_ports_from_json):
+        with transaction.atomic():
+            for port in Port.objects.all().only('name', 'active'):
+                if port.name.lower() not in all_ports_from_json:
                     port.active = False
                     port.save()
 
--- a/app/ports/templates/ports/ajax-filters/filtered_table.html
+++ b/app/ports/templates/ports/ajax-filters/filtered_table.html
@@ -23,7 +23,7 @@
         </thead>
         {% for port in ports %}
             <tr>
-                <td><a href="{% url 'port_detail' port.port.name %}">{{ port.port.name }}</a></td>
+                <td><a {% if port.active is False %}class="text-danger"{% endif %} href="{% url 'port_detail' port.port.name %}">{{ port.port.name }}</a></td>
                 <td>{{ port.port.version }}</td>
                 <td>{{ port.port.description }}</td>
             </tr>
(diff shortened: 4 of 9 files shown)


arjunsalyan:
This has been deployed on the test site:
http://ec2-52-34-234-111.us-west-2.compute.amazonaws.com/

mojca:
This PR contains three somewhat unrelated commits. I'll try to shortly comment on each one of them.

* Looping through all ports to identify old deleted ports

    I need to take a closer look, but in the future I would be in favour of perhaps switching to updating port info "after each commit" (but maybe not more frequently than every N minutes in case that a large number of updates flows in). Generally this is only needed as a one-time fix for something that was broken for a while, and ideally we'll never need it again. If we now execute this code every few commits, even if it only runs for 15 seconds, such code snippets might keep adding up to very long execution times. It does make sense to execute some cleanups on regular basis, but doing it each time still sounds a bit of an overkill in the long run.
    We will also need to address licences one day (#44) which would also require modifying all ports, despite most of them not being actually updated (I would vote for modifying the string in the json file).
    So it would probably make sense to come up with some more generic way of batch-updating ports when such a need arises. Any thoughts about that?

* Show only active ports by default in port listings

    * Yes, definitely, this needs to go in. Ideally we would also add some unit tests though, to make sure that all those queries are correct. They are by now a bit easy too easy to break if someone else starts contributing code.
    * We still have one bug left: if you check the page of `apache-ant`, it lists `jump` as having a build dependency on `apache-ant`. But `jump` was deleted, so we shouldn't be listing that dependency. I see three options: (1) delete deleted ports, (2) delete all dependencies whenever a port gets deleted, (3) make complex queries to ensure that only active dependencies are listed. I think I would go for (2), and probably also not display the dependency fields for any deleted ports. I guess that still leaves the question about whether we should keep maintainers. Probably we should remove maintainers as well?

* Add checkbox to toggle displaying of deleted ports

    * Actually, it's the other commit that addresses #110 :) The problem was that deleted ports are shown in the first place. (Don't consider this to be a serious feedback.)
    * You mentioned that you deployed the solution, but you should at least mention some deleted ports on which this could be tested. I searched the logs and now found `python37-devel` and `jump` as two recent examples.
    * I'm not ever sure that we really really need this checkbox, but if we do have it, I would probably call it "Include delete ports" rather than "Show deleted ports". The latter might suggest that only deleted ports will be shown.
    * I wanted to say that when the ports are shown as search results, we would probably need some kind of visual marker for deleted, but I see that you already added that, thanks a lot. Do we want something similar for obsolete ports as well?
    * Generally I would say that similar functionality could be used for other things, but I'm pretty sure we don't want do include deleted ports when searching through various categories, maintainers etc. If/when we add some advanced search functionality, I would probably even delete the huge "Include deleted ports" checkbox from the main page and only leave the functionality on the "very special search" page.

umeshksingla:
> You mentioned that you deployed the solution, but you should at least mention some deleted ports on which this could be tested. I searched the logs and now found python37-devel and jump as two recent examples.

I felt that too when I tried to check and but eventually felt asleep looking for deleted ports in macports-ports repository but I tried thinking about it in a different way. If he would have given an example of a deleted port, we might have just checked that and merged and would not be able to catch any potential bugs (any slightly complicated example like `jump` you just mentioned) which we may find with a little exploration.

> I'm not ever sure that we really really need this checkbox, but if we do have it, I would probably call it "Include delete ports" rather than "Show deleted ports". The latter might suggest that only deleted ports will be shown.

Yes, "Include" is certainly better than "show". I wanted to go for something like:
Status: [ ] Active      [ ] Deleted      [ ] Obsolete      [ ] xyz

But it seemed too much for now. And I am sure people who are using it will find it very inconvenient to go through the huge number of results the search returns.

> I would probably even delete the huge "Include deleted ports" checkbox from the main page and only leave the functionality on the "very special search" page.

Completely agree. Or search bar can support queries with tags like `include:deleted python-devel3`, much like Gmail or Google Search.

mojca:
> I wanted to go for something like:
> Status: [ ] Active [ ] Deleted [ ] Obsolete [ ] xyz
> 
> But it seemed too much for now.

No way. Already the checkbox for deleted ports might be too much.

Also note that our list of deleted ports is very incomplete (almost semi-random). If I search for deleted ports, I might eventually expect to find them all.

> Or search bar can support queries with tags like `include:deleted python-devel3`, much like Gmail or Google Search.

Maybe, but that's still quite far down the road of all other features that we might want to have.

arjunsalyan:
>  Generally this is only needed as a one-time fix for something that was broken for a while, and ideally we'll never need it again. If we now execute this code every few commits, even if it only runs for 15 seconds, such code snippets might keep adding up to very long execution times.

So you are suggesting that we go for the old method?

>   * We still have one bug left: if you check the page of `apache-ant`, it lists `jump` as having a build dependency on `apache-ant`. But `jump` was deleted, so we shouldn't be listing that dependency.

I did not want to change any information we picked from the Portfile. And I feel that we should not, but maybe I am wrong.

>   * You mentioned that you deployed the solution, but you should at least mention some deleted ports on which this could be tested.

Oh, sorry, my bad.

>   Do we want something similar for obsolete ports as well?

Maybe yellow for obsolete ports?

>   If/when we add some advanced search functionality, I would probably even delete the huge "Include deleted ports" checkbox from the main page and only leave the functionality on the "very special search" page.

Port discovery is one of the most important part of the app, why not convert the homepage to a full fledged search page? The box that contains the search box could become more advanced? I am just saying.

mojca:
> > Generally this is only needed as a one-time fix for something that was broken for a while, and ideally we'll never need it again. If we now execute this code every few commits, even if it only runs for 15 seconds, such code snippets might keep adding up to very long execution times.
> 
> So you are suggesting that we go for the old method?

What was the old method?

I'm trying to brainstorm here, pointing out some concerns (reasons why doing this too often might not be the best idea).

What would you or @umeshksingla consider the best solution here?

> > * We still have one bug left: if you check the page of `apache-ant`, it lists `jump` as having a build dependency on `apache-ant`. But `jump` was deleted, so we shouldn't be listing that dependency.
> 
> I did not want to change any information we picked from the Portfile. And I feel that we should not, but maybe I am wrong.

It might be ok to show `apache-ant` as a dependency of `jump` (on `jump`'s page), but not to show `jump` as a dependent of `apache-ant`. When developers make changes to a port like `apache-ant` they might want to check if all dependent ports still work, or check if the port can safely be renamed or deleted without affecting others. Including deleted ports on that list is just contributing to noise.

Keeping deleted ports around is useful to some extent (if users want to create a new port it could help to know that such a port existed in the past; external links to that port would not break etc.). But the exact information about that port is of very limited use. If the port was first obsoleted, most of the information was gone at that point anyway. There is definitely no maintainer (even if one was present before the deletion, the port is de-facto no longer maintainer), dependencies are bogus, website might be broken, ...

You could certainly keep the information about dependencies, but I'm afraid this will needlessly complicate proper handling at the other end, where you really don't want to show deleted dependents / dependencies.

> > Do we want something similar for obsolete ports as well?
> 
> Maybe yellow for obsolete ports?

Sounds fine. (Or basically any other way, I don't really care.)

> > If/when we add some advanced search functionality, I would probably even delete the huge "Include deleted ports" checkbox from the main page and only leave the functionality on the "very special search" page.
> 
> Port discovery is one of the most important part of the app, why not convert the homepage to a full fledged search page? The box that contains the search box could become more advanced? I am just saying.

What if we start by creating a dedicated `/search` page and experimenting on that one?

What we could do on the first/main page is then add a `More / Advanced` button which shows additional search options (it could be either javascript or redirect to a different search page; probably the first). Then "Include deleted ports" could be one of the advanced search options.

arjunsalyan:
> What was the old method?

We were taking all the ports under the updated portdirs and checking if they were available in the latest JSON.

> I'm trying to brainstorm here, pointing out some concerns (reasons why doing this too often might not be the best idea).

Initially, even I did not like the idea of running this every time. But seeing the speed and simplicity of the code, I got inclined towards this. However, as you mentioned, running this every time might not be a good idea.

> What would you or @umeshksingla consider the best solution here?

In any case, we will have to run it at least once. Maybe, we can merge this and later switch back to the old method?

> It might be ok to show `apache-ant` as a dependency of `jump` (on `jump`'s page), but not to show `jump` as a dependent of `apache-ant`.

Okay.

> What we could do on the first/main page is then add a `More / Advanced` button which shows additional search options (it could be either javascript or redirect to a different search page; probably the first). Then "Include deleted ports" could be one of the advanced search options.

Seems good!

arjunsalyan:
One thing I am realising is, I need to break it into several different Pull Requests, otherwise one thing is just getting delayed due to other.

mojca:
Yes, breaking into smaller chunks probably makes sense.

Also one random thought. I provided some feedback about the imperfection of importing the new builders, but on the other hand ... it will do less harm to have suboptimal queries (but complete build history in the database) than not show any builds at all. After all it's just about displaying data, so we could easily go with your initial approach, just import data, and work on improving the sql queries later.

arjunsalyan:
> Also one random thought. I provided some feedback about the imperfection of importing the new builders, but on the other hand ... it will do less harm to have suboptimal queries (but complete build history in the database) than not show any builds at all. After all it's just about displaying data, so we could easily go with your initial approach, just import data, and work on improving the sql queries later.

Yes, that was what I also though initially, but now I have both things in place: https://github.com/macports/macports-webapp/pull/119 . However, if we quickly just want to add those new builders, I can create a new PR.

arjunsalyan:
The newer update and search mechanism handles everything out-of-the-box. This PR is obsolete.

PR #125, opened 2019-09-27
Fix 'Most installed version' on port installation stats

### LINKED ISSUE #123: Incorrect ‘most installed version’ is reported.
Each port seems to give ‘user(s)’ as the most installed version.... see e.g.

https://ports.macports.org/port/clang-8.0/stats?days=30&days_ago=0

arjunsalyan:
Closes: https://github.com/macports/macports-webapp/issues/123

### DIFF
--- a/app/ports/templates/ports/port-detail/installation_stats.html
+++ b/app/ports/templates/ports/port-detail/installation_stats.html
@@ -37,7 +37,7 @@
                 </tr>
                 <tr>
                     <th scope="row">Most Installed Version</th>
-                    <td><strong>{{ port_installations_by_port_version.first.version }}</strong> : {{ port_installations_by_port_version.first.num }} user(s)</td>
+                    <td>{% most_installed_version port_installations_by_port_version %}</td>
                 </tr>
             </table>
     <div class="row">
--- a/app/ports/templatetags/utilities.py
+++ b/app/ports/templatetags/utilities.py
@@ -9,3 +9,10 @@ def find_count_for_version(distribution, version):
         if i['version'] == version:
             return i['num']
     return 0
+
+
+@register.simple_tag
+def most_installed_version(distribution):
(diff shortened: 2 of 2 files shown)


arjunsalyan:
I have tested this and it is a minor change, so I am merging this PR as the functionality is broken.

app/ports/templates/ports/port-detail/installation_stats.html
```diff
@@ -37,7 +37,7 @@ <h4 class="border-bottom">Statistics: <span style="font-size:12px"><button type=
                 </tr>
                 <tr>
                     <th scope="row">Most Installed Version</th>
-                    <td><strong>{{ port_installations_by_port_version.first.version }}</strong> : {{ port_installations_by_port_version.first.num }} user(s)</td>
```
umeshksingla, line 40: Why did this stop working in the first place?

arjunsalyan: There were changes done to the `port_installations_by_port_version` query set:
1. Versions are no longer sorted by number of users, so first object will not give the most installed version.
2. It is no longer a query set, but a list of dictionaries.

mojca:
Thanks for fixing this.

But the sole fact that this got broken without anyone noticing is a strong suggestion that we need an unit test!!!

umeshksingla:
I still don't understand why this broke in the first place. I tried going through the commits where it could have last changed but couldn't find anything. Maybe I need to take a deeper look.

mojca:
#116?

umeshksingla:
Oh. I skipped that latest commit because I felt why would `port_installations_by_port_version` be affected by something related to visual ordering ever. Turns out the commit was made to fix something else which in turn fixed x-axis sorting by itself.

This was stupid of me.

PR #127, opened 2019-10-10
Show only active ports by default in port listings

arjunsalyan:
I am trying to break https://github.com/macports/macports-webapp/pull/122 so that the bug can be fixed without being delayed due to the "Include deleted ports" filter.


mojca:
When we implement more general filters we could probably improve / generalize the code a bit, but it looks fine for now.

PR #128, opened 2019-10-10
Add function to loop over all ports to identify deleted ports

arjunsalyan:
This is another part of https://github.com/macports/macports-webapp/pull/122 .

We need to run this function at least once so that the ports that were missed to be marked as deleted get marked correctly. However, I am not sure of the right way to do this.

Any suggestions: @mojca @umeshksingla

### DIFF
--- a/app/ports/models/port.py
+++ b/app/ports/models/port.py
@@ -283,6 +283,24 @@ class Port(models.Model):
                     port.active = False
                     port.save()
 
+    # Run through all the ports to identify deleted ports
+    @classmethod
+    def full_deleted_ports_run(cls):
+        all_ports_from_json = set()
+        data = Port.PortIndexUpdateHandler().sync_and_open_file()
+
+        # Generate a set of all the portname available in the JSON file
+        for port in data['ports']:
+            portname = port['name'].lower()
+            all_ports_from_json.add(portname)
+
+        # Loop over all the ports in the database and check if it is available in all_ports_from_json
+        with transaction.atomic():
+            for port in Port.objects.all().only('name', 'active'):
+                if port.name.lower() not in all_ports_from_json:
+                    port.active = False
+                    port.save()
+
     class PortIndexUpdateHandler:
         @staticmethod
         def sync_and_open_file():


app/ports/models/port.py
```diff
@@ -283,6 +283,24 @@ def mark_deleted(cls, dict_of_portdirs_with_ports):
                     port.active = False
                     port.save()
 
+    # Run through all the ports to identify deleted ports
+    @classmethod
+    def full_deleted_ports_run(cls):
+        all_ports_from_json = set()
+        data = Port.PortIndexUpdateHandler().sync_and_open_file()
```
mojca, line 290: I need to check again, but does running this function always fetch the latest version of the json file?

I kind of feel that the functionality should be separate from rsync-ing the file.

app/ports/models/port.py
```diff
@@ -283,6 +283,24 @@ def mark_deleted(cls, dict_of_portdirs_with_ports):
...
+
+        # Generate a set of all the portname available in the JSON file
+        for port in data['ports']:
+            portname = port['name'].lower()
+            all_ports_from_json.add(portname)
+
+        # Loop over all the ports in the database and check if it is available in all_ports_from_json
+        with transaction.atomic():
+            for port in Port.objects.all().only('name', 'active'):
```
mojca, line 299: I would additionally filter out non-active ports. Once you end up with 80% of the ports being deleted, this will make a significant difference.

mojca (commented):
Looks ok to me. I guess that another step is needed to actually trigger this, right?

I left some notes, and tests cannot hurt, but I also don't mind if it goes in without modifications, just to get the database fixed.

app/ports/models/port.py
arjunsalyan: Yes, the function `sync_and_open_file` first fetches the latest file and then reads it.

arjunsalyan:
> Looks ok to me. I guess that another step is needed to actually trigger this, right?

Yes, and I am not sure how do we trigger this. Maybe a migration?

app/ports/models/port.py
arjunsalyan: Right, thanks. Very fair point.

mojca:
> > Looks ok to me. I guess that another step is needed to actually trigger this, right?
> 
> Yes, and I am not sure how do we trigger this. Maybe a migration?

Hmmm ... actually migration sounds like a suitable idea (we do need to fix the database, after all). Maybe add a comment next to the function saying that we had to fix the situation once due to buggy code, but that we might want to run that code every now and then. We don't immediately need the perfect solution, and having a function available + running it once might do for now, and then we can think about the best ways to potentially do full updates.

I do imagine cases like when we change the format of json to create more user-friendly licence fields. In such cases we'll need to run through the whole database as well.

Bottomline: we can merge this before implementing migration (or together, if ready), and worry about the best way to run this later.

arjunsalyan:
We can merge this together with the required migration file. I will add the migration file after we merge https://github.com/macports/macports-webapp/pull/119 just to avoid any conflicts in the migrations.

arjunsalyan:
I am unable to call this model method from the migrations. Seems like we are not allowed to do this unless I move all the code inside the migration file itself.

arjunsalyan:
The new update mechanism has an improved version of this. This PR can be closed.

PR #130, opened 2019-10-10
Fix ports-list styling of maintainer's page for smaller screens

arjunsalyan:
On smaller screens the maintainer's page currently looks like this:

Screenshot 2019-10-10 at 11 37 14 PM

After applying this commit, it will look like this:

Screenshot 2019-10-10 at 11 38 02 PM


mojca:
Thanks.

PR #132, opened 2019-10-28
Add builder for macOS 10.15

mojca:
Not tested yet ...

### DIFF
--- a/app/ports/migrations/0012_add_10.15_worker.py
+++ b/app/ports/migrations/0012_add_10.15_worker.py
@@ -1,0 +1,17 @@
+from django.db import migrations
+
+
+def add_new_builder(apps, schema_editor):
+    Builder = apps.get_model('ports', 'Builder')
+    Builder.objects.create(name='10.15_x86_64', display_name='10.15')
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('ports', '0011_populate_display_name.py'),
+    ]
+
+    operations = [
+        migrations.RunPython(add_new_builder),
+    ]
--- a/app/ports/tests/test_update_portinfo.py
+++ b/app/ports/tests/test_update_portinfo.py
@@ -15,7 +15,13 @@ class TestUpdatePortinfo(TransactionTestCase):
 
     def test_updates(self):
         call_command('update-portinfo', '3c6f37828d091670e7b9a2676757b2e468ee3d52', 'cb7086953124c73ffb616e955653176e3c4be02c')
-        self.assertEquals(Port.objects.count(), 14)
+        try:
+            port1 = Port.objects.get(name='youtube-dl')
+            port2 = Port.objects.get(name='py-yapf')
+            update_success = True
+        except Port.DoesNotExist:
+            update_success = False
+        self.assertEquals(update_success, True)
 
     def test_deleted(self):
         Port.mark_deleted({


mojca:
We are getting some pretty unusual unit test failures which we might want to fix:
```
Failed to append AtomicParsley as a dependency to youtube-dl. Not Found.
Failed to append ffmpeg as a dependency to youtube-dl. Not Found.
Failed to append python37 as a dependency to youtube-dl. Not Found.
Failed to append zip as a dependency to youtube-dl. Not Found.
Failed to append python37 as a dependency to youtube-dl. Not Found.
Failed to append python27 as a dependency to py27-yapf. Not Found.
Failed to append py27-setuptools as a dependency to py27-yapf. Not Found.
Failed to append yapf_select as a dependency to py27-yapf. Not Found.
Failed to append python35 as a dependency to py35-yapf. Not Found.
Failed to append py35-setuptools as a dependency to py35-yapf. Not Found.
Failed to append yapf_select as a dependency to py35-yapf. Not Found.
Failed to append python36 as a dependency to py36-yapf. Not Found.
Failed to append py36-setuptools as a dependency to py36-yapf. Not Found.
Failed to append yapf_select as a dependency to py36-yapf. Not Found.
Failed to append python37 as a dependency to py37-yapf. Not Found.
Failed to append py37-setuptools as a dependency to py37-yapf. Not Found.
Failed to append yapf_select as a dependency to py37-yapf. Not Found.
Failed to append python38 as a dependency to py38-yapf. Not Found.
Failed to append py38-setuptools as a dependency to py38-yapf. Not Found.
Failed to append yapf_select as a dependency to py38-yapf. Not Found.
F.................
======================================================================
FAIL: test_updates (ports.tests.test_update_portinfo.TestUpdatePortinfo)
----------------------------------------------------------------------
  File "/home/travis/build/macports/macports-webapp/app/ports/tests/test_update_portinfo.py", line 18, in test_updates
    self.assertEquals(Port.objects.count(), 14)
AssertionError: 15 != 14
```

arjunsalyan:
This test depends on macports-ports repository. It takes two fixed commits and carries out updates. Probably the commits I had chosen have been modified, but they were from months ago. I haven’t been able to take a thorough look, I will do this as soon as I have access to my computer.

ryandesign:
Our repositories are configured to disallow modifications to past commits to master or release branches.

mojca:
What has been more likely to happen is that the ports themselves were modified to a sufficient extent (and we are using the portindex from the server rather than generating it ourselves based on the given commit) that the test no longer passes as a consequence. I would need to take a closer look anyway.

arjunsalyan:
The reason is: a new subport has been added to the port `py-yapf`, and hence the count of updated ports has increased by 1.

I will change the test to not count the ports but look for the parent port only. But then in future if any one of the ports py-yapf or YouTube-dl is deleted then the test will again start failing.

arjunsalyan:
I have fixed the issue of the failing test, and this PR looks good to merged.

mojca:
I forced pushed to the branch, but the Travil doesn't want to restart the tests.

We may still want to see if we can further improve the tests, but I agree that it's important to do two things fast:
* Add the 10.15 builder
* Make sure that full history gets fetched for all new builders.

Maybe we should add an environmental variable to the deployed version of our app (just this one instance) as a flag that certain tasks only need to be done there and not on copies done by users for the sake of testing.

arjunsalyan:
I believe we can merge this after increasing the number of builds to be fetched. It is currently 20, but we can increase it as the number of builds for 10.15 would be much higher, while we decide on getting the entire build history.

umeshksingla:
Do you mean we change it to 100 or 200 or something and then change it back to 20? I am good at it and don't think we ll have any problems.

arjunsalyan:
> Do you mean we change it to 100 or 200 or something and then change it back to 20? I am good at it and don't think we ll have any problems.

Yes, but even without that, the PR looks good to be merged.

umeshksingla:
@arjunsalyan I guess we can ignore codecov report for a migration file? Or maybe you want to take a look.

ryandesign:
> Do you mean we change it to 100 or 200 or something and then change it back to 20?

Why wouldn't you want to get the information from all of the > 4000 builds that have happened so far on the 10.15 builder?

arjunsalyan:
> Why wouldn't you want to get the information from all of the > 4000 builds that have happened so far on the 10.15 builder?

We can make it 4000 or more. But this would mean 4000+ continuous requests to the server. If we are good with that, then we can make it 4000 and try to fill that gap some other way.

ryandesign:
I'm not familiar with the webapp code any why there is a limit at all, but I think the point is that we want build status information for as many ports as we can, so by all means please retrieve all the available info from the build server.

ryandesign:
For the 10.15 builder, I would expect that you would start getting build information at build 1 and continue until the latest build, and then get incremental information later.

If increasing whatever this limit is to 4000 will also affect the other builders—for example if that would retrieve older builds for the new 10.6-10.8 builders than you have thus far retrieved—then I hope that it will correctly deal with the fact that builds happen chronologically. We would not want information about an old build of a port to overwrite information about a newer build of that port that had been previously retrieved. Each of the new 10.6-10.8 builders has fewer than 10,000 builds so far, which is how far back we keep build information, so you could go back to the beginning on those as well. But don't discard existing build information for 10.6-10.8, since we are only rebuilding ports that used libstdc++; builds of non-C++ software remain as they were before the new builders came online.

arjunsalyan:
> For the 10.15 builder, I would expect that you would start getting build information at build 1 and continue until the latest build, and then get incremental information later.

I am seeing that builds start at number 0? Or is it 1?
This makes it easier to fetch all the builds.

> If increasing whatever this limit is to 4000 will also affect the other builders—for example if that would retrieve older builds for the new 10.6-10.8 builders than you have thus far retrieved—then I hope that it will correctly deal with the fact that builds happen chronologically. We would not want information about an old build of a port to overwrite information about a newer build of that port that had been previously retrieved. Each of the new 10.6-10.8 builders has fewer than 10,000 builds so far, which is how far back we keep build information, so you could go back to the beginning on those as well. But don't discard existing build information for 10.6-10.8, since we are only rebuilding ports that used libstdc++; builds of non-C++ software remain as they were before the new builders came online.

Currently, this will not affect other builders. The app first checks if it has any previous builds for a builder, if it does have- then it fetches only the new builds since the last fetch, otherwise it simply fetches 20 last builds.
For 10.15, I am going to make changes such that instead of fetching the last 20 builds, we start at 0.

ryandesign:
> I am seeing that builds start at number 0? Or is it 1?

Oh yes, you are right, [the first build is 0](https://build.macports.org/builders/ports-10.15_x86_64-builder/builds/0).

mojca:
Still not sure if you don't need to enter last build -1 to fetch build 0 first.

We should probably still do something to ensure that fetching from 0 won't happen on test deployments.

app/ports/migrations/0012_add_10.15_worker.py
```diff
@@ -0,0 +1,17 @@
...
+
+def add_new_builder(apps, schema_editor):
+    Builder = apps.get_model('ports', 'Builder')
+    Builder.objects.create(name='10.15_x86_64', display_name='10.15')
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('ports', '0011_populate_display_name.py'),
```
arjunsalyan, line 12: This migration did not get applied. After some testing I found that adding `.py` to the last migration's name causes it to fail, so I believe it to be the most probable reason.

We can try to fix this by removing this `.py` but I am spectacle about it, as it did not work for my development setup and I had to make a new migration.

arjunsalyan: @neverpanic must have received an email regarding the failure.

neverpanic: Haven't noticed. Maybe I missed it, though.

PR #134, opened 2019-11-04
Modify test dependent on macports-ports

arjunsalyan:


arjunsalyan:
I am merging this now, we need to get 10.15 builder history quickly.

PR #139, opened 2019-11-16
Fetch builds from number 0 for new builders

arjunsalyan:
This is a temporary change to allow fetching all builds for 10.15 (to be added- https://github.com/macports/macports-webapp/pull/132). Similar change can be applied to fetch all builds for 10.6-10.8.

Ideally, we need to have a flag which would give the flexibility to allow fetching from beginning or just some previous builds.

### DIFF
--- a/app/ports/models/buildhistory.py
+++ b/app/ports/models/buildhistory.py
@@ -106,7 +106,9 @@ class BuildHistory(models.Model):
             if build_number_loaded:
                 build_in_database = build_number_loaded[0].build_id + 1
             else:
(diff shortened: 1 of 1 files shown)


ryandesign:
> Similar change can be applied to fetch all builds for 10.6-10.8.

The 10.7 builder will probably reach build number 10,000 within a day or two, at which point old builds will start getting deleted. So if you want to fetch all of them, I would start with 10.7 soon. 10.6 x86_64 will be the next soonest to reach 10,000, followed by 10.8 and lastly 10.6 i386.

PR #194, opened 2020-03-05
Remove border around port detail

### LINKED ISSUE #145: Remove box and border around port name / description
On port pages, there is the port name at the top, with a short description and port health indicators below. These are all enclosed by a box with a medium gray border and a light gray background. I would suggest removing the box and the background and just letting white space do the job.

me-diru:
Fixes #145 
Have checked locally
Before :
image

After:
image

### DIFF
--- a/app/ports/templates/ports/portdetail.html
+++ b/app/ports/templates/ports/portdetail.html
@@ -44,7 +44,7 @@
     {% endif %}
     <!-- Deleted port warning end-->
 
-        <div class="rounded bg-light border border-secondary" style="padding: 8px">
+        <div style="padding: 8px">
         <h4>{{ req_port.name }}</h4>
         <p class="text-secondary">{{ req_port.description }}</p>
         {% if req_port.active is True %}


mojca:
Thank you very much for your contribution.

May I please ask you to include a screenshot, so that other developers who don't run a local copy will be able to provide feedback?

me-diru:
Have included the screenshots

arjunsalyan (approved):
Thank you very much for addressing the issue(s), and especially for setting up the development environment locally.

The PR looks good if we are ready to adopt the new layout as it is.

arjunsalyan:
Oh, I just found that we need to improve the commit message.

It should be "Remove border around port detail", instead of "removed". The commit message could also include the Issue it is referring to, in this case, something like :

```
Remove border around port detail

Closes: https://github.com/macports/macports-webapp/issues/145
```

This link might be useful which @mojca had shared earlier https://chris.beams.io/posts/git-commit/

arjunsalyan:
This is no longer relevant owing to a major redesign of the port page.

PR #202, opened 2020-04-13
Restore the number of builds fetched to last N when no previous builds are available

arjunsalyan:
When the app finds that for a particular builder there are no builds available in the database, it used to fetch only the last N builds, N is defined by `BUILDS_FETCHED_COUNT` in `Macports.config`. But we changed it to fetch all the builds temporarily for the special case of `10.15` builder.

This can now be restored.

### DIFF
--- a/app/ports/models/buildhistory.py
+++ b/app/ports/models/buildhistory.py
@@ -106,9 +106,9 @@ class BuildHistory(models.Model):
             if build_number_loaded:
                 build_in_database = build_number_loaded[0].build_id + 1
             else:
-                # build_in_database = last_build_number - BUILDS_FETCHED_COUNT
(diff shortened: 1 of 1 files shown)


arjunsalyan:
This is a very small change I thought would be important to be done before we start working on refactoring.

@mojca : I can go ahead and merge it, but if you have something to add?

mojca:
Feel free to merge, it looks ok to me.


We can think of better ways to handle this in the future, but no need to worry about it now.
I can imagine a special variable to tell Django whether it should treat the installation as a development/testing version or as an officially deployed version. In the officially deployed version it could always try to go as far back as possible, while it would never fetch more than N commits for everyone else.

But again: just pull it as it is.

PR #203, opened 2020-04-13
Add index to BuildHistory table

arjunsalyan:
This index is missing on the BuildHistory table. It might be the reason behind the slow response time of the port health feature.


mojca (approved):

app/ports/migrations/0013_buildhistory_add_index.py
```diff
@@ -0,0 +1,17 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('ports', '0012_add_10_15_worker'),
+    ]
+
+    operations = [
+        migrations.AddIndex(
+            model_name='buildhistory',
+            index=models.Index(fields=['port_name', 'builder_name', '-time_start'], name='ports_build_port_na_5f3089_idx'),
```
mojca, line 15: Just to clarify: so this index already exists in the model file, it was just forgotten to be added as a migration?

arjunsalyan: Yes, I checked the file history which revealed that no migration was generated after adding this index to the model. Also when I ran the command locally, this migration was generated.

Issue #149, opened 2019-12-02
Copyable search URL


ryandesign:
When I type into the search box, a search automatically happens without pressing Return, however even if I do press Return, the URL doesn't change. This makes it impossible to copy and paste the URL of the search results, which I might want to do for example in reply to a message on the mailing list where someone has asked if we have a particular port.

arjunsalyan:
I had started work on a separate and more advanced search page but could not take it forward. I think the search on home should be left as it is- just a quick way to discover ports. And an advanced version could shift to a separate page.

PR #206, opened 2020-04-21
Add support for search and search-by params on port search page

mems:
Ex: https://ports.macports.org/ports/?search=django&search_by=name

Plus:

- prefilled input with provided values
- update the URL when the search is made or if search is cleared
- remove params if search is empty

Note: this use [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL) and [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) APIs that aren't support by IE5-11 but all others browsers and [`history.pushState()` and `history.replaceState()`](https://developer.mozilla.org/en-US/docs/Web/API/History) that aren't supported by IE5-9

### DIFF
--- a/app/static/js/ajax-search.js
+++ b/app/static/js/ajax-search.js
@@ -3,12 +3,15 @@ var requestTimer;
 function ajaxCallSearch() {
     var data = {};
     var currentSearchRequest = null;
-    data[$("input[name=search_by]:checked").val()] = $('#search').val();
+    var search = $('#search').val();
+    data[$("input[name=search_by]:checked").val()] = search;
     data['csrfmiddlewaretoken'] = $("input[name=csrfmiddlewaretoken]").val();
     data['search_by'] = $("input[name=search_by]:checked").val();
-    data['search_text'] = $('#search').val()
+    data['search_text'] = search;
+    
+    updateLocation();
 
-    if ($('#search').val()) {
+    if (search) {
         $('#filtered_table').show();
         $('#main-content').hide();
         currentSearchRequest = $.ajax({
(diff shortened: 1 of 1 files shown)


arjunsalyan (approved):
This looks really good. I have tested the changes and it works as accepted. Thank You

mojca:
@mems: Thank you for your contribution.
Which `search_by` parameters are allowed?

arjunsalyan:
> @mems: Thank you for your contribution.
> Which `search_by` parameters are allowed?

The search backend only supports `name` and `description`. These changes are just at the client-side and hence are restricted by what is supported by the backend.

Having said that, in its current state, this change is suited only for two `search_by` parameters.

app/static/js/ajax-search.js
```diff
@@ -33,12 +36,27 @@ function ajaxCallSearch() {
...
-        clearTimeout(requestTimer);
-        requestTimer = setTimeout(function () {
-            ajaxCallSearch();
-        }, 1000);
-    });
+    var params = new URLSearchParams(location.search.substring(1));
+    var search = params.get('search');
+    var defaultSearchBy = $("input[name=search_by]:checked").val();
+    var otherSearchBy = $("input[name=search_by]:not(:checked)").val();
+    var searchBy = params.get('search_by') === defaultSearchBy ? defaultSearchBy : otherSearchBy;
```
arjunsalyan, line 43: We should be able to support more `search_by` parameters.

One thing I can think of is to generate an array of all the `search_by` parameters and validate with that, instead of just checking with a single default. This might be helpful if we plan to add more `search_by` params in the future.

mems: I do that only for security purpose.

To support multiple values, we could use instead:

```js
var availableSearchByValues = $("input[name=search_by]").get().map(function (el) { return el.value });
var searchBy = params.get('search_by');
if (availableSearchByValues.indexOf(searchBy) < 0) {
    searchBy = $("input[name=search_by]:checked").val();
}
```

arjunsalyan: This looks better, it would allow for more flexibility. What possible security flaw do you see with it?

mems: It have a limited impact here, but [never trust user input](https://blog.sheddow.xyz/css-timing-attack/):

https://github.com/macports/macports-webapp/blob/8b5d68b08038a869ad54b62d87e8d17a3eb89f3a/app/static/js/ajax-search.js#L53

arjunsalyan: I totally agree with you, but it should not be a problem here. However, I am fine with the current implementation also.

arjunsalyan:
This addresses the issue: https://github.com/macports/macports-webapp/issues/149

PR #208, opened 2020-04-25
Improve port health database query

arjunsalyan:
It is not sure by when we complete the refactoring of the app, or when the new code gets deployed. In the current deployed version, the port page takes quite a time to load and it is due to this query.

I have tried to improve it and tested it locally on my small dataset.


arjunsalyan:
@verm : can't request a review, so mentioning.

PR #215, opened 2020-05-14
Convert _ to . while sorting port versions

arjunsalyan:
This fixes strange version sorting for cases:

`1.18_0` should be ordered below `1.20.1_1`, but removing the underscore was giving opposite results. Instead of removing the underscores, they can be converted to dots (.).


mojca:
It would be really nice to cover the functionality with some simple unit test.

(Another issue is epoch change, but that's almost a different topic.)

PR #223, opened 2020-05-23
Add various filters to search page

arjunsalyan:
I have changed some UI behaviours, the search page is now more of an “Explore ports” page. The one on home still helps quick go to ports by using port name, while the search page gives some more powerful “search by” and “filter by” options.

Demo: http://macports.silentfox.tech/search/?selected_facets=maintainers_exact:mojca&livecheck_outdated=on

In the demo, we are searching by a maintainer and then selecting only the outdated ports.

Todo:

- [x] Filter ports with no maintainer
- [x] Facets for maintainers, categories and variants


rajgoesout:
The search is looking great!
Does [port summary page](http://macports.silentfox.tech/port/sord/summary) need to have a search bar? Or perhaps you could use the new search bar in that page.

arjunsalyan:
> The search is looking great!
> Does [port summary page](http://macports.silentfox.tech/port/sord/summary) need to have a search bar? Or perhaps you could use the new search bar in that page.

Right, thanks. I am planning to add the same search as on the homepage to the navigation bar, do you have any other suggestions?

mojca:
Awesome!

I need to check, but I have a feeling that some results are missing (in my case at least some of geant4, clhep, gate, ...)

Is it possible to search for unmaintained ports?

arjunsalyan:
> I need to check, but I have a feeling that some results are missing (in my case at least some of geant4, clhep, gate, ...)

Yes, some are missing. It is because the livecheck I ran after updating the table schema could not finish. Searching is being done based on the `has_updates` column, and it was added later, so it has not yet populated for all the ports. I will run another full livecheck tonight.

**UPDATE**: I ran full livecheck again, do you now see all ports? geant4, clhep and gate are there now.

> Is it possible to search for unmaintained ports?

Not yet, but will add it soon, it is in the Todo also.

**UPDATE**: Added.

arjunsalyan:
This PR is now complete from the functionalities point of view. Once we merge this, I will start replacing variants, categories and maintainers pages with respective facets on the search page (that would be a lot of code simplification and reduction).

PR #224, opened 2020-05-25
Switch to same version sorting algorithm as used by macports-base

### LINKED ISSUE #216: Build stats. pages hang whilst loading.
See this for a number of ports. build stats. page just hangs with the spinning waiting symbol. see e.g.

https://ports.macports.org/port/libgcc10/stats?days=30&days_ago=0

arjunsalyan:
fixes: https://github.com/macports/macports-webapp/issues/216

Python's `LooseVersion` started to fail with a `TypeError`, so we need to switch to a different version sorting algorithm.

I have tried to implement the same algorithm as from [here](https://github.com/macports/macports-base/blob/master/src/pextlib1.0/vercomp.c) which seems to be the RPM version sorting algorithm.

Also, added a few test cases

### DIFF
--- a/app/ports/utilities/sort_by_version.py
+++ b/app/ports/utilities/sort_by_version.py
@@ -1,13 +1,59 @@
-from distutils.version import LooseVersion
+import re
+from functools import cmp_to_key
 
 
-def cleaned_version(version):
-    return version.replace("_", ".").replace("-", ".")
+# Tries to apply the same version sorting algorithm as used by macports-base
+# https://github.com/macports/macports-base/blob/master/src/pextlib1.0/vercomp.c
+def version_compare(version1, version2):
+    # Split on all non-alphanumeric characters and delete any empty ('') elements
+    # that might occur due to consecutive non-alphanumerics
+    version1_segments = list(filter(None, re.split('[^a-zA-Z0-9]', version1)))
+    version2_segments = list(filter(None, re.split('[^a-zA-Z0-9]', version2)))
+
+    version1_len = len(version1_segments)
+    version2_len = len(version2_segments)
+
+    # Now traverse the two lists and compare segments at each index
+    i = 0
+    while i < version1_len and i < version2_len:
+        s1 = version1_segments[i]
+        s2 = version2_segments[i]
+
+        # If version1's segment is a numeric segment and version2's segment
+        # is non-numeric, consider version1 to be a newer version
+        if s1.isnumeric() and (not s2.isnumeric()):
+            return -1
+
+        # If the segments are of different types, consider version2 be newer
+        if (s1.isnumeric() and s2.isalpha()) or (s1.isalpha() and s2.isnumeric()):
+            return 1
+
+        # The segments are either both numeric or both alphabetical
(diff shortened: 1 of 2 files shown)


jmroot (approved):
Oh, wow, you reimplemented it in python. I was thinking more along the lines of making vercomp.c into an extension module. That would have the advantage of the exact same code being used in both places, but this has other advantages from keeping it pure python.

Looks good to me. I don't suppose you'd like to add the same test cases to base? 😀

arjunsalyan:
> Looks good to me. I don't suppose you'd like to add the same test cases to base? 😀

Are you suggesting that the test cases could be added to the base? I can do that.

mojca:
Thanks!

app/ports/utilities/sort_by_version.py
```diff
@@ -1,13 +1,59 @@
...
+        s1 = version1_segments[i]
+        s2 = version2_segments[i]
+
+        # If version1's segment is a numeric segment and version2's segment
+        # is non-numeric, consider version1 to be a newer version
+        if s1.isnumeric() and (not s2.isnumeric()):
+            return -1
+
+        # If the segments are of different types, consider version2 be newer
+        if (s1.isnumeric() and s2.isalpha()) or (s1.isalpha() and s2.isnumeric()):
```
mojca, line 28: @jmroot: Oh, are we really using such a strange code? Shouldn't the one that's numeric (or the one that's not) at least consistently win?

app/ports/utilities/sort_by_version.py
```diff
@@ -1,13 +1,59 @@
...
+    # that might occur due to consecutive non-alphanumerics
+    version1_segments = list(filter(None, re.split('[^a-zA-Z0-9]', version1)))
+    version2_segments = list(filter(None, re.split('[^a-zA-Z0-9]', version2)))
+
+    version1_len = len(version1_segments)
+    version2_len = len(version2_segments)
+
+    # Now traverse the two lists and compare segments at each index
+    i = 0
+    while i < version1_len and i < version2_len:
```
mojca, line 18: An alternative way (no need to change anything) could have been
```python
for i in range(min(version1_len, version2_len)):
```

app/ports/utilities/sort_by_version.py
arjunsalyan: https://github.com/macports/macports-base/blob/08d87705e9cce7b1f64c061d4335695be44099f6/src/pextlib1.0/vercomp.c#L89

I had the same thought, but I could interpret it this way only.

The case when version1 is numeric and other is not has been covered before. So, I guess the numeric wins anyways. But one case that is not covered is, when version2 is numeric and version1 is alphanumeric (not alphabetic).

app/ports/utilities/sort_by_version.py
arjunsalyan: Hmm...thanks, that would have been a nicer way.

jmroot:
> Are you suggesting that the test cases could be added to the base? I can do that.

That would give more assurance that the behaviour is the same. Thanks.

PR #230, opened 2020-06-12
Refresh port info update mechanism and other general upgrades

arjunsalyan:


app/port/models.py
```diff
@@ -48,151 +48,33 @@ def get_absolute_url(self):
...
-                        categories.add(Category(name=category))
-                except KeyError:
-                    pass
-            batch = list(categories)
-            Category.objects.bulk_create(batch)
-            return
-
-        port_id_map = {}
+    def add_or_update(cls, data):
+        from category.models import Category
```
rajgoesout: Is there any performance gain by importing inside the function? If not, then I would keep all the imports at the top https://www.python.org/dev/peps/pep-0008/

rajgoesout (commented):
LGTM

app/port/models.py
arjunsalyan: I did this to prevent cyclic imports. However, I will try if I can get them at the top, if there are no other conflicts.

rajgoesout: You could use `import foo as bar` if there is a name clash/circular import.

**PS.** this isn't super important

arjunsalyan: Thank you, I had fixed the cyclic imports earlier itself but forgot to change the import statements. This has been fixed.

PR #238, opened 2020-07-17
Experimental dark mode

arjunsalyan:


rajgoesout:
This looks cool!
The background color of the search box that says 'start typing' changes inconsistently

arjunsalyan:
> This looks cool!
> The background color of the search box that says 'start typing' changes inconsistently

That is because of the typeahead layer on top of the search box. Reloading the page will fix that.

PR #254, opened 2020-09-05
Port.update(): Improve MultipleObjectsReturned err msg

neverpanic:
The cronjob is currently failing with

```
ERROR OUTPUT:
  File "/code/app/manage.py", line 15, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.6/dist-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.6/dist-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.6/dist-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.6/dist-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
  File "/code/app/ports/management/commands/update-portinfo.py", line 55, in handle
    Port.update(ports_to_be_updated_json)
  File "/code/app/ports/models/port.py", line 292, in update
    run_updates(data)
  File "/code/app/ports/models/port.py", line 277, in run_updates
    full_update_dependencies(ports)
  File "/code/app/ports/models/port.py", line 262, in full_update_dependencies
    dependencies.append(Port.objects.get(name__iexact=depends.rsplit(':', 1)[-1]))
  File "/usr/local/lib/python3.6/dist-packages/django/db/models/manager.py", line 82, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/django/db/models/query.py", line 412, in get
    (self.model._meta.object_name, num)
ports.models.port.MultipleObjectsReturned: get() returned more than one Port -- it returned 2!
```

Improve the error message printed when this exception occurs to find out why this seems to be a problem. Note that this started showing up on August 31st, shortly after the django update to 2.2.13 was merged in ecfcbb7da3aa79dce124f3e5f2b9bfc73e341db3.

### DIFF
--- a/app/ports/models/port.py
+++ b/app/ports/models/port.py
@@ -264,6 +264,12 @@ class Port(models.Model):
                                     print("Failed to append {} as a dependency to {}. Not Found.".format(
                                         depends.rsplit(':', 1)[-1],
                                         port['name']))
+                                except MultipleObjectsReturned as e:
+                                    print("Could not append {} as a dependency to {}, because the depspec does"
+                                          " not uniquely identify a port".format(
+                                        depends,
+                                        port['name']))
+                                    raise e
                             obj.save()
                             obj.dependencies.add(*dependencies)
 


neverpanic:
Ignoring the reduced code coverage for now, we can always revert this change later once we understood why it happens.

arjunsalyan:
Thank you, identifying the port for which duplicate entries got created has to be the first step to fix this.

The reason why we ended up with two entries seems to be this get_or_create() method https://github.com/macports/macports-webapp/blob/ecfcbb7da3aa79dce124f3e5f2b9bfc73e341db3/app/ports/models/port.py#L184

Probably some inconsistency in the case of port name lead to two rows being created, with same port name but different case.

I addressed this exact situation in the newer version of the app https://github.com/macports/macports-webapp/blob/01efdfda9de59c1905bf077f8cc0089f1b0df2db/app/port/models.py#L119 . But adding a `unique` constraint over port name is also needed.

neverpanic:
I checked the database and could not find a duplicate, though. Let's see what the next cron run will tell us.

neverpanic:
Port in question is `libexosip2`. I'm assuming this may be related to https://github.com/macports/macports-ports/commit/f5c9781df6c76686f3505d24d4aa3ffa7d6e9d82 which changed the case of the name.

We have both variants:
```
webapp=> select id, version, name from ports_port where name ILIKE 'libexosip2';
  id   | version |    name
-------+---------+------------
  5464 | 5.1.1   | libeXosip2
 26333 | 5.1.3   | libexosip2
(2 rows)
```

Should I just delete the outdated one?

neverpanic:
See also #255, my code didn't work.

arjunsalyan:
Yes, the outdated one can be deleted, it anyways would have got overwritten if not for the inconsistency.

I will submit patches so that this does not happen again. But the exception would also be helpful, so we can keep it in place.

neverpanic:
I have deleted the duplicate entry. I'll re-run the cron to ensure this fixed the issue.

neverpanic:
Issue fixed, thanks for your help.

PR #255, opened 2020-09-05
Port.update(): Fix exception name

neverpanic:
`MultipleObjectsReturned` does not exist, but `Port.MultipleObjectsReturned` does. Since this could happen again the next time somebody case-renames a port, let's keep the code in to avoid additional debugging the next time.


arjunsalyan:
Thank you, this will be helpful and should be kept in place. I was wondering that this exception is under the model class, but did not check, sorry.

neverpanic:
Still no tests, but this is still strictly better than the broken version I merged before and that I had to manually fix on the server after deployment, so let's merge this.

PR #260, opened 2021-02-05
Change port detail page title to include "MacOS"

gagarine:
The page do not mention MacOS. This could improve SEO a bit.

### DIFF
--- a/app/ports/templates/ports/portdetail.html
+++ b/app/ports/templates/ports/portdetail.html
@@ -9,7 +9,7 @@
     <script type="text/javascript" src="https://www.gstatic.com/charts/loader.js"></script>
 {% endblock %}
 
-{% block title %}Install {{ req_port.name }} Port on MacOS -  |{% endblock %}
+{% block title %}Install {{ req_port.name }} Port on macOS -  |{% endblock %}
 
 {% block content %}
     <br>
--- a/app/ports/templates/ports/portdetail.html
+++ b/app/ports/templates/ports/portdetail.html
@@ -9,7 +9,7 @@
     <script type="text/javascript" src="https://www.gstatic.com/charts/loader.js"></script>
 {% endblock %}
 
-{% block title %}{{ req_port.name }}- Port |{% endblock %}
+{% block title %}Install {{ req_port.name }} Port on MacOS -  |{% endblock %}
 
 {% block content %}
     <br>


gagarine:
In adition I believe it should contain the install command for the specific port, but I didn't find a nice way to add it in the current design.

Screenshot 2021-02-05 at 14 24 43

It should also have clear and big link back to main macport website. Today if someone arrive on a port details without knowing what is macport, he will certainly be lost.

app/ports/templates/ports/portdetail.html
```diff
@@ -9,7 +9,7 @@
     <script type="text/javascript" src="https://www.gstatic.com/charts/loader.js"></script>
 {% endblock %}
 
-{% block title %}{{ req_port.name }}- Port |{% endblock %}
+{% block title %}Install {{ req_port.name }} Port on MacOS -  |{% endblock %}
```
neverpanic:
```suggestion
{% block title %}Install {{ req_port.name }} Port on macOS -  |{% endblock %}
```
Apple's official capitalization of "macOS" starts with a lower case "m". Other than that, this lg2m.

neverpanic (changes requested):

gagarine:
The build seem to fail. But it seem to be a problem out of this commit.

neverpanic (approved):
Change lg2m, let me look at the build failure, though – this won't get deployed if the build fails.

neverpanic:
Seems like this needs updated python dependencies. Let me check what I can do about those.

neverpanic:
See also #261. CI also needs fixing, maybe I'll find time to take a look at this later or in the coming days.

neverpanic:
So now the title tag is `Install htop Port on macOS - | MacPorts`. That's not really ideal.

Shouldn't it rather be something like `Install htop on macOS | MacPorts`?

gagarine:
> Shouldn't it rather be something like Install htop on macOS | MacPorts?

Yes. It would be cool. I didn’t propose this because I taught you may want to keep that it is a port.  But with MacPorts as the seconde part of the url I think it’s ok. 

What about `Install htop on macOS | MacPorts Package Manager` ?

 In addition adding the install commande would be great so people can copy and past and new user can see how easy it is.

neverpanic:
I think people google "install xyz macOS", not necessarily "install xyz macOS package manager", so I don't think we need the "package manager" addition. It would be great if we could drop the dash before the pipe, though, that looks really out of place.

PR #262, opened 2021-02-14
Remove "ports" from page title

gagarine:
https://github.com/macports/macports-webapp/pull/260  Follow-up


neverpanic (approved):
Thanks! The server will auto-deploy after merge.

Issue #273, opened 2021-04-13
Only 10.15 builds are being imported into web app


ryandesign:
I am curious why the web app only shows build information for one build (on 10.15) for this port:

https://ports.macports.org/port/lexbor/summary

Builds were successful on all OS versions. This is a new port that no other port depends on, so its build information should have been readily available as a separate build on all OS versions.

See commit status on this commit: https://github.com/macports/macports-ports/commit/b5b22439d06dc135b77031083933bd5962b9cc93

ryandesign:
It looks like only build information for 10.15 is being imported, for all ports:

https://ports.macports.org/ports/all_builds/

Any idea why?

neverpanic:
We're at least not getting any error messages from the cronjob. Anything else I'll have to investigate first :/

ryandesign:
According to the build server web logs, Braeburn continues to retrieve build information for all platforms, e.g.:
```
136.243.18.213 - - [19/Apr/2021:20:50:03 +0000] "GET /json/builders/ports-10.11_x86_64-builder/builds/-1 HTTP/1.1" 200 7176 "-" "Python-urllib/3.8" "-"
136.243.18.213 - - [19/Apr/2021:20:50:04 +0000] "GET /json/builders/ports-10.11_x86_64-builder/builds/134087 HTTP/1.1" 404 153 "-" "Python-urllib/3.8" "-"
136.243.18.213 - - [19/Apr/2021:20:50:05 +0000] "GET /json/builders/ports-10.12_x86_64-builder/builds/-1 HTTP/1.1" 200 7175 "-" "Python-urllib/3.8" "-"
136.243.18.213 - - [19/Apr/2021:20:50:06 +0000] "GET /json/builders/ports-10.12_x86_64-builder/builds/139159 HTTP/1.1" 404 153 "-" "Python-urllib/3.8" "-"
136.243.18.213 - - [19/Apr/2021:20:50:07 +0000] "GET /json/builders/ports-10.13_x86_64-builder/builds/-1 HTTP/1.1" 200 6328 "-" "Python-urllib/3.8" "-"
136.243.18.213 - - [19/Apr/2021:20:50:08 +0000] "GET /json/builders/ports-10.13_x86_64-builder/builds/103815 HTTP/1.1" 404 153 "-" "Python-urllib/3.8" "-"
[...]

PR #274, opened 2021-04-20
Improve handling of errors while fetching build history

arjunsalyan:
https://github.com/macports/macports-webapp/issues/273

This skips the current build if 404 response code is recieved from buildbot, otherwise it stops the process for the builder so that a retry attempt is made in the next run of the cronjob.

### DIFF
--- a/app/ports/models/buildhistory.py
+++ b/app/ports/models/buildhistory.py
@@ -51,9 +51,12 @@ class BuildHistory(models.Model):
             try:
                 with urllib.request.urlopen(url, context=gcontext) as u:
                     data = json.loads(u.read().decode())
-                return data
+                    resp_code = u.getcode()
+                return data, resp_code
+            except urllib.error.HTTPError as err:
+                return {}, err.code
             except urllib.error.URLError:
-                return {}
+                return {}, None
 
         def get_build_properties(array):
             properties = {}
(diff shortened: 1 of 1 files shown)


mojca (approved):
This looks ok to me.

It maybe wouldn't hurt to have a simple test, but it's already good as it is.

app/ports/models/buildhistory.py
```diff
@@ -99,7 +102,9 @@ def load_database(data):
 
         for buildername in builders:
             # fetch the last build first in order to figure out its number
-            last_build_data = get_data_from_url(get_url_json(buildername, -1))
+            last_build_data, resp_code = get_data_from_url(get_url_json(buildername, -1))
```
neverpanic, line 105:
```suggestion
            last_build_data, _ = get_data_from_url(get_url_json(buildername, -1))
```
It's customary to use `_` for unused variables in Python.

neverpanic (approved):
A nitpick, lg2m otherwise.

app/ports/models/buildhistory.py
ryandesign: I tried to commit this suggestion but got "Invalid email for web commit. Refresh and try again." If anyone is able to commit this suggestion please do, then let's squash and merge the PR.

mojca: I'm sorry, I misread the contents and for some reason thought that this has already been fixed (while it was still just a suggestion). Anyway, I merged this now, we can still do this cosmetic change in a separate commit, hoping that the site will be fixed now.

PR #287, opened 2021-07-17
Skip 0th and 12th month from monthly stats

arjunsalyan:
Since, we collect stats many times in a month, the result of 0th and 12th month show incomplete data. This happens because we are still collecting data for the 12th month (which is actually the ongoing month) and the 0th month comes into picture due to how we calculate `last 12 months`.  0th month shows data for just last 1 or 2 days of the month, which we anyways do not intend to show in the charts.

Screenshot 2021-07-17 at 2 11 23 PM

### DIFF
--- a/app/stats/serializers.py
+++ b/app/stats/serializers.py
@@ -9,6 +9,7 @@ from stats.models import Submission
 from stats.utilities.sort_by_version import sort_list_of_dicts_by_version
 from stats.validators import ALLOWED_DAYS_FOR_STATS as ALLOWED_DAYS
 from stats.validators import ALLOWED_PROPERTIES, ALLOWED_GENERAL_PROPERTIES
+from stats.utilities import dates
 
 
 class PortStatisticsSerializer(serializers.Serializer):
@@ -120,11 +121,10 @@ class PortMonthlyInstallationsSerializer(serializers.Serializer):
         is_valid = self.validate_context()
         if not is_valid:
             return PortInstallation.objects.none()
-        today_day = datetime.datetime.now().day
-        last_12_months = datetime.datetime.now(tz=datetime.timezone.utc) - datetime.timedelta(days=int(today_day) + 365)
+
         result = PortInstallation.objects \
             .only('id') \
-            .filter(port__iexact=self.port_name, submission__timestamp__gte=last_12_months) \
+            .filter(port__iexact=self.port_name, submission__timestamp__gte=dates.get_first_day_of_month_x_months_ago(12)) \
             .annotate(datetime=TruncMonth('submission__timestamp')) \
             .order_by('datetime') \
             .annotate(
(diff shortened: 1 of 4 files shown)


mojca:
The fix sounds ok, I'm just wondering whether we could perhaps change the timestamp of the first submission instead.
I admit that I never even noticed the problem with the first month. (I saw it on the last month of course, but I took that as "part of the game" if your month isn't complete yet.)

I've seen some pages use a bit of a shaded background or dashed lines for the last month with incomplete data, but I'm not sure whether or not we should make things overcomplicated.

I definitely agree that the first month should go (I would only try to change the python expression for the start date: I can look at that). I'm not sure what to do with the last month.

app/stats/serializers.py
```diff
@@ -121,7 +121,9 @@ def get_result(self, obj):
         if not is_valid:
             return PortInstallation.objects.none()
         today_day = datetime.datetime.now().day
-        last_12_months = datetime.datetime.now(tz=datetime.timezone.utc) - datetime.timedelta(days=int(today_day) + 365)
+        last_12_months = datetime.datetime.now(tz=datetime.timezone.utc) - datetime.timedelta(365)
+        last_12_months.replace(day=1)
```
mojca: But what about `hour=0, minute=0, second=0, microseconds=0`?
Just thinking loud about alternatives ... For example:
```python
today = datetime.datetime.utcnow().date()
first_day_of_the_month_one_year_ago = datetime.datetime(today.year - 1, today.month, 1, tzinfo=datetime.timezone.utc)
```
That should take care of both leap years and UTC time zone (at least if I understand the first one correctly).

The code to retrieve this weird timestamp is needed twice and sounds complex enough that we could in principle write a simple python function in one of the "common utilities" places and call that function here rather than duplicating the code.

Also, "last_12_months" implies an interval. We need a variable name that implies a fixed timestamp. Anything shorter and clearer than my suggestion would probably be welcome (`one_year_ago` is not accurate, but it might do the job).

(Note: `today_day` is no longer needed.)

arjunsalyan: I still feel the first approach to be more flexible, and more useful if we want to make a utility function. Mostly because we can supply the number of months we want to go back, and then get the first day of that month.

Thank you for pointing out misleading variable names and `hour=0, minute=0, second=0, milliseconds=0`?

mojca: How many days back will you go in utility function that says "go back for 1 month"?

(Some suggest using `dateutil` for this kind of date manipulations.)

arjunsalyan: Oh, sorry I missed to add the use of `monthDelta` https://pythonhosted.org/MonthDelta/ .

Once this we go back certain months ago in time, and then replace `day=1`.

mojca: Using `monthDelta` correctly definitely sounds better than manually subtracting 365 days.

PR #288, opened 2021-07-17
Do not show "Total submissions" on stats page

arjunsalyan:
I have setup stats data older than 18 months to be purged. So "total submissions" will not provide accurate information. I also feel that this wasn't useful anyways.

Screenshot 2021-07-17 at 5 53 05 PM

### DIFF
--- a/app/stats/templates/stats/stats.html
+++ b/app/stats/templates/stats/stats.html
@@ -27,10 +27,10 @@
             <br>
             <table class="table table-striped">
                 <tbody>
-                <tr>
+                <!-- <tr>
                     <th scope="row">Total submissions:</th>
                     <td class="text-right">{{ total_submissions }}</td>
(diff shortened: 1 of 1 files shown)


mojca:
I'm neither in favour of purging the old statistics nor in favour of deleting this.
Basically, this information is really helpful when trying to interpret the data about port popularity.

arjunsalyan:
> I'm neither in favour of purging the old statistics nor in favour of deleting this.
> Basically, this information is really helpful when trying to interpret the data about port popularity.

Okay, we can disable the purge then. But I only imported the submissions received after 1st Jan, 2020 (total user count will not be affected). They are there in the backup, I will add them to the database as well later.

The only reason why I thought of it was some 18 million rows in the installed ports table. And 17 million of those have been received after 1st jan, 2020.

PR #303, opened 2021-07-21
Missing map files

mojca:
My browser is throwing some errors in inspector:
```
[Error] Failed to load resource: the server responded with a status of 404 (Not Found) (https://ports.macports.org/static/css/bootstrap.min.css.map, line 0)
[Error] Failed to load resource: the server responded with a status of 404 (Not Found) (https://ports.macports.org/static/css/darktheme.css.map, line 0)
[Error] Failed to load resource: the server responded with a status of 404 (Not Found) (https://ports.macports.org/static/js/bootstrap.min.js.map, line 0)
```
I have no clue whether this is the correct fix, I just added two files from the same bootstrap version that's included in the page, and removed the `map` for most likely non-existent `darktheme.css.map`.


arjunsalyan (approved):

arjunsalyan:
Other way is to simply remove the references for the map files, either of them is fine! Thank You

mojca:
By removing the references we would be modifying the original file. Not that this is a serious problem, but ...

PR #319, opened 2021-11-02
Ticket query: Allow comma as word separator

Dave-Allured:
Bug fix.  Support comma in addition to space, as word separator between multiple port names.

Example before bug fix:
https://trac.macports.org/report/16?max=1000&PORT=(%5E%7C%5Cs)libheif($%7C%5Cs)

After bug fix:
https://trac.macports.org/report/16?max=1000&PORT=(%5E%7C%5Cs%7C,)libheif($%7C%5Cs%7C,)

### DIFF
--- a/app/port/views.py
+++ b/app/port/views.py
@@ -150,7 +150,7 @@ def default_port_page_toggle(request, name):
 @cache_page(60 * 60)
 def port_tickets(request, name):
     port_name = request.GET.get('port_name')
-    URL = "https://trac.macports.org/report/16?max=1000&PORT=(%5E%7C%5Cs){}($%7C%5Cs)".format(port_name)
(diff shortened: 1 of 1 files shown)


arjunsalyan:
Thank you. This looks good.

Dave-Allured:
Great.  For a more interesting example, plug "gcc11" or some other port name with many tickets into those two URL's.

https://trac.macports.org/report/16?max=1000&PORT=(%5E%7C%5Cs)gcc11($%7C%5Cs)
https://trac.macports.org/report/16?max=1000&PORT=(%5E%7C%5Cs%7C,)gcc11($%7C%5Cs%7C,)

Also compare with the original behavior of the "PORT" dialog box on the Trac Report no. 16 query page.  In essence, the original no. 16 query matches by substring.  The webapp matches only full port names, which IMO is most appropriate for this webapp.  You can see that the results are different from above, in this example:

https://trac.macports.org/report/16?max=1000&PORT=gcc11

PR #327, opened 2021-12-20
Port health: Use background loading

arjunsalyan:
This improves performance for the port details page because port-health is an expensive query.


mascguy:
Arjun, thanks again for all of the work you've done - and continue to do - for the site. Great stuff!

arjunsalyan:
While this has improved the loading time of port-details page, port health is still taking ages to load. That query needs to be improved.

PR #332, opened 2022-01-18
Hide port long description with full stop

harens:
Currently, if `description == long_description`, the long description is hidden.

image

It's quite common in some ports for the long description to simply be `{*}${description}.` (note the full stop), in which case the long description is shown.

image

This very simple PR aims to change this behaviour to hide the long description in both cases.


mojca (approved):
Thanks a lot for the contribution. It makes perfect sense to me.

arjunsalyan (approved):

PR #338, opened 2022-03-23
Comma-delimit numbers; use tabular figures in tables

amake:
I find large numbers difficult to read without delimiters. Also, in tables, using [tabular-nums](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-numeric) can help a lot with legibility.

I did try to test this with a local setup, but I'm not very confident that I didn't maybe break something.

### DIFF
--- a/app/port/templates/port/includes/port-header.html
+++ b/app/port/templates/port/includes/port-header.html
@@ -27,7 +27,7 @@
     <span class="float-right">
         {% if is_followed %}<a href="{% url 'unfollow_port' port.name %}" class="btn btn-sm btn-light"><i class="fa fa-minus-circle"></i> Unfollow</a>
         {% else %}<a href="{% url 'follow_port' port.name %}" class="btn btn-sm btn-light"><i class="fa fa-plus"></i> Add to my watchlist</a>{% endif %}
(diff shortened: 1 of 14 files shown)


arjunsalyan:
Thank you. From a quick look it all looks good to me, but I do not have access to my computer at this moment, so can't verify the actual behaviour. Let's keep it open until someone or I can verify that this does not break anything.

amake:
Using #339 I was able to locally confirm that all pages I could feasibly test are working without issue.

The only one I wasn't able to test was the profile screen which requires logging in.

mojca:
Can you please share some screenshots here?

amake:
Before

image

After

image

amake:
The screenshots above illustrate 99% of the change. Basically anywhere an integer is shown it will be comma-delimited.

The tabular number thing is extremely subtle: without changing the font, *only* numbers become monospaced. This helps with tabular data where you want to be able to visually compare the sizes of numbers; otherwise for instance a run of `1`s will make a number appear much narrower than a perhaps larger number that has multiple `8`s.

Because I don't have mpstats working locally, I had to manually "fake" the following "after" picture:

Before

Screen Shot 2022-05-06 at 21 46 54

After

Screen Shot 2022-05-06 at 21 46 42

PR #419, opened 2025-11-07
New modern and decoupled UI

arjunsalyan:
New nextjs based UI which is decoupled from the django server. It utilises the APIs which were already being exposed by the server.

Try it out before we release it: https://macports-new.arjunsalyan.com (there might be some glitches and missing pieces)

This is 100% AI generated code.


nilason:
Looks nice and is really responsive! I couldn't find live check though.

Issue #135, opened 2019-11-08
Port health is not tied to port version/revision


ryandesign:
The port health indicators are misleading because they are not tied to the port version/revision.

For example, mkvtoolnix was recently updated to 39.0.0. At the time of filing this ticket, the [port health indicators](https://ports.macports.org/port/mkvtoolnix/summary) show green for 10.14, 10.13, 10.12, red for 10.11, 10.10, 10.9, green for 10.8 & 10.7, and red for 10.6 and 10.6 i386.

But if you drill down, you discover that the statuses for 10.9 and up are for the 39.0.0_0 build while the 10.8 and earlier statuses are for the previous 30.1.0_1 build. The 10.8 and earlier builders are very busy right now rebuilding hundreds of ports for the libc++ transition and have not yet gotten around to building mkvtoolnix 39.0.0.

It seems like the status indicators for each builder need to be cleared when the port version or revision is updated. And when you get status information from the buildbot, you need to ensure that it is for the current version/revision of the port.

mojca:
Yes, I'm painfully aware of this and I have put this on the task list (not entered in the tracker though).

The prerequisite to get this working is to make the version information easily accessible from the buildbot history, and something that you would need to deploy on our buildbot master. The only  other alternative is to parse full build logs from all builds, but that will likely download plenty gigabytes from your server.

Options would be:
* quick hack to `mpbb` to print the version information to the "statistics" file (next to file size / build dir size / time to build)
* same or similar step as the one which prints port name that would also print port version (ideal)
* for older builds the only available option would be to parse full logs, but that would ideally need to be done from your local network, else we'll clog the server

mojca:
Also, once we do this, we might need to somewhat extend the port health, so that it will show the health of the last build even when the newest build isn't yet ready (but make it clear that it's an older status).

ryandesign:
> * same or similar step as the one which prints port name that would also print port version (ideal)

Printing the full `name @version_revision+variants` spec in the builder step instead of just the name seems like a reasonable thing to do. Would that be ok?

ryandesign:
> * for older builds the only available option would be to parse full logs, but that would ideally need to be done from your local network, else we'll clog the server

What format would be best? Just a csv file for each builder with build number, port name, version, revision, variants?

[...]

PR #438, opened 2026-08-14
Include port version and revision in build history

AMDmi3:
I've recently launched a [service](https://github.com/AMDmi3/wiyci) which aggregates package build logs and parses warnings and other issues from these, hopefully extending upstream awareness and encouraging fix contributions, currently hosted at https://wiyci.repology.org. I'd like to add macports as another build logs source, for that I need to be able to get to a build log for a given package/version/revision. I plan to do it using build history API (complete algorithm described in the [issue](https://github.com/AMDmi3/wiyci/issues/31) under Macports section), however the history currently lacks version and revision fields so it's not possible to pick a build for the right version. To fix that, I propose to add these fields. Or maybe there's a simpler solution I have missed?

For the record, the service is designed not to generate excess load on the server, limiting request rate by 1 request in 3 seconds by default. If macports support is added it'll have to catch up by downloading logs for all covered projects (~10k as of now, but I hope to grow it few times) which should be 10k * number of covered macos versions (which can be limited with only the latest version if so desired), after which it'll only download logs for updated versions, which, judging by already supported repositories, is less than 100 per day.

### DIFF
--- a/app/buildhistory/serializers.py
+++ b/app/buildhistory/serializers.py
@@ -14,7 +14,7 @@ class BuildHistorySerializer(serializers.ModelSerializer):
 
     class Meta:
         model = BuildHistory
-        fields = ('port_name', 'builder_name', 'build_id', 'status', 'time_start', 'time_elapsed', 'watcher_id')
+        fields = ('port_name', 'port_version', 'port_revision', 'builder_name', 'build_id', 'status', 'time_start', 'time_elapsed', 'watcher_id')
 
 
 class FilesSerializer(serializers.ModelSerializer):


jmroot:
I don't think this can be correct in general, because the port version in a build is not necessarily the same as the latest version known to the web app when it finds out about the build. This is the same basic problem as in #135.

AMDmi3:
> I don't think this can be correct in general, because the port version in a build is not necessarily the same as the latest version known to the web app when it finds out about the build

If I understood that correctly, that shouldn't be a problem for my case, as long as the version tied to the build is the version actually built. wiyci discovers version from Repology (which is the current one in Macports), then tries to fetch builds for that version. It is ok and in fact expected that there are no builds with that version at the given moment, it retries a few times with the exponentially increasing period (1 day, then 2 days after that, then 4 days after that, then 8 days after that), so if the build appears with some delay it will not be missed. The purpose of this PR is to be able to tell which builds are for which version, and thus be able to skip builds not related to the desired version.

jmroot:
> The purpose of this PR is to be able to tell which builds are for which version, and thus be able to skip builds not related to the desired version.

I understand your goal (and accomplishing it would be great), but the problem is that no mapping between port versions and builds currently exists. When fetching new builds from the buildbot, there is simply no indication in the metadata of which version of the port was built. As discussed in #135, that is fixable, but short of parsing logs to extract the information, it requires a change on the buildbot side as well.

AMDmi3:
Oh, I understand now. That's a pity.

